PR #14, opened 2017-07-14
Configurable perf metrics

phijor:


src/perf/event_provider.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+static bool event_is_openable(const platform::CounterDescription& ev)
+{
+    struct perf_event_attr attr;
+    memset(&attr, 0, sizeof(attr));
+    attr.size = sizeof(attr);
+    attr.type = ev.type;
+    attr.config = ev.config;
+    attr.config1 = ev.config1;
+    attr.exclude_kernel = lo2s::config().exclude_kernel;
```
tilsche: Always enable `exclude_kernel` for the check. Later code will fallback to `exclude_kernel=1` regardless of the `config().exclude_kernel`.

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,384 @@
...
+    const auto slash_pos = ev_desc.find_first_of('/');
+    if (slash_pos == npos)
+    {
+        throw EventProvider::InvalidEvent("missing '/' in event description");
+    }
+
+    const std::string pmu_name = ev_desc.substr(0, slash_pos);
+    const std::string event_name = ev_desc.substr(slash_pos + 1);
+
+    // PMU -- performance measurement unit
```
tilsche: performance *monitoring* unit

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,384 @@
...
+    {
+        event.config1 |= apply_mask(value, mask);
+    }
+}
+
+const platform::CounterDescription sysfs_read_event(const std::string& ev_desc)
+{
+    namespace fs = boost::filesystem;
+    static constexpr auto npos = std::string::npos;
+    const auto slash_pos = ev_desc.find_first_of('/');
```
tilsche: perf requires a trailing `/`, we should at least allow that. It might be helpful to require it for disambiguation.

src/trace/counters.cpp
```diff
@@ -37,14 +40,34 @@ Counters::Counters(pid_t pid, pid_t tid, Trace& trace_, otf2::definition::metric
...
+    for (const auto& ev : user_events)
+    {
+        try
+        {
+            const auto event_desc = perf::EventProvider::get_event_by_name(ev);
+            counters_.emplace_back(tid, event_desc.type, event_desc.config, event_desc.config1);
+        }
+        catch (const perf::EventProvider::InvalidEvent& e)
+        {
+            lo2s::Log::warn() << "'" << ev << "' does not name a known event, ignoring!";
```
tilsche: `e.what()` might be more helpful than "does not name a known event".

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,384 @@
...
+    int fd = syscall(__NR_perf_event_open, &attr, 0, -1, -1, 0);
+    if (fd == -1)
+    {
+        switch (errno)
+        {
+        case ENOTSUP:
+            Log::warn() << "perf event not supported by the running kernel: " << ev.name;
+            break;
+        default:
+            Log::warn() << "perf event not available: " << ev.name;
```
tilsche: I think this is too common on healthy systems, a reduction to `debug` is probably in order. When requested, this will trigger higher level warnings anyway.

include/lo2s/perf/event_provider.hpp
```diff
@@ -0,0 +1,96 @@
...
+class EventProvider
+{
+public:
+    using EventMap = std::unordered_map<std::string, platform::CounterDescription>;
+
+    EventProvider();
+    EventProvider(const EventProvider&) = delete;
+    void operator=(const EventProvider&) = delete;
+
+    static const inline EventProvider& instance()
```
bmario: Please note, according to the C++ standard, class member functions _defined_ inside of a class definition are implicitly inline. (§ 9.3.2 [class.mfct])

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,389 @@
...
+    {
+        WHOLE_MATCH,
+        BIT_BEGIN,
+        BIT_END,
+    };
+
+    std::uint64_t mask = 0x0;
+
+    static const std::regex bit_mask_regex(R"((\d+)?(?:-(\d+)))");
+    static const std::sregex_iterator end;
```
bmario: Is the std::sregex_iterator that expensive?
Otherwise I'd change it to not being static.

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,389 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with lo2s.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <lo2s/config.hpp>
+#include <lo2s/log.hpp>
+#include <lo2s/perf/event_provider.hpp>
+
+#include <boost/filesystem.hpp>
```
bmario: Add the missing `#include <boost/filesystem/fstream.hpp>`. I don't know, why @tilsche didn't add them in the first place in other files, as he's the one with IncludeWhatYouUse 😕

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,389 @@
...
+                                     const std::string& format)
+{
+    static constexpr auto npos = std::string::npos;
+    const auto colon = format.find_first_of(':');
+    if (colon == npos)
+    {
+        throw EventProvider::InvalidEvent("invalid format description: missing colon");
+    }
+
+    const std::string target_config = format.substr(0, colon);
```
bmario: Why not also const auto?

src/trace/counters.cpp
```diff
@@ -58,21 +83,34 @@ Counters::Counters(pid_t pid, pid_t tid, Trace& trace_, otf2::definition::metric
 otf2::definition::metric_class Counters::get_metric_class(Trace& trace_)
 {
     auto c = trace_.metric_class();
+    const auto& user_events = lo2s::config().perf_events;
 
-    for (const auto& description : platform::get_mem_events())
+    for (const auto& ev : lo2s::config().perf_events)
```
bmario: `for (const auto& ev : user_events)`

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,389 @@
...
+    PERF_EVENT_HW("cache-references", CACHE_REFERENCES),
+    PERF_EVENT_HW("cache-misses", CACHE_MISSES),
+    PERF_EVENT_HW("branch-instructions", BRANCH_INSTRUCTIONS),
+    PERF_EVENT_HW("branch-misses", BRANCH_MISSES),
+    PERF_EVENT_HW("bus-cycles", BUS_CYCLES),
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 0, 0)
+    PERF_EVENT_HW("stalled-cycles-frontend", STALLED_CYCLES_FRONTEND),
+    PERF_EVENT_HW("stalled-cycles-backend", STALLED_CYCLES_BACKEND),
+#endif
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 3, 0)
```
phijor: These checks are only there because they guard macros that aren't defined on lower kernel versions. If they weren't guarded, `lo2s` would fail to compile on those kernels.

Nonetheless, every event is checked at runtime to be available. If you compiled on a higher kernel version where these events were defined in a header and ran on a lower one, you'd get a `debug`-level warning (see [`event_is_openable`](https://github.com/tud-zih-energy/lo2s/blob/04912b89fd63c4f0a951c13fb0f465c933645006/src/perf/event_provider.cpp#L126)).

src/perf/event_provider.cpp
phijor: Whoops... Yep, that should probably not be `static`.

src/config.cpp
```diff
@@ -73,16 +97,16 @@ void parse_program_options(int argc, const char** argv)
...
+        ("verbose,v", po::value(&verbosity)->zero_tokens(),
+             "verbose output (specify multiple times to get increasingly more verbose output)")
         ("mmap-pages,m", po::value(&config.mmap_pages)->default_value(16),
              "number of pages to be used by each internal buffer")
         ("readout-interval,i", po::value(&read_interval_ms)->default_value(100),
              "time interval between metric and sampling buffer readouts in milliseconds")
-        ("raw-tracepoint-event,e", po::value(&config.tracepoint_events),
+        ("tracepoint,t", po::value(&config.tracepoint_events),
              "enable global recording of a raw tracepoint event (usually requires root)")
+        ("event,e", po::value(&config.perf_events),
```
tilsche: Yes, this is a good moment to break with the options towards compatibility with `perf record`.

src/perf/event_provider.cpp
tilsche: Maybe I got stuck, IWYU can be a pain. But why didn't our CI detect it in the other file in the first place ;-).

bmario: I have [no](https://github.com/tud-zih-energy/lo2s/commit/8cf83799c4b570178909181a9e58a7fa2fe873de) [idea](https://github.com/tud-zih-energy/lo2s/commit/80346e9eccad1dbd4d68f66036a5c4b819177166)

src/config.cpp
rschoene: @tilsche uhm, how about no?

tilsche: Is there anything out there that gets broken by this change that we can't easily fix? I'd really love to have `lo2s` as a drop-in replacement for `perf record` for like 80% of use cases.

rschoene: @tilsche valid point, ok.

src/config.cpp
```diff
@@ -157,23 +181,34 @@ void parse_program_options(int argc, const char** argv)
...
         lo2s::logging::set_min_severity_level(nitro::log::severity_level::warn);
     }
     else
     {
-        lo2s::logging::set_min_severity_level(nitro::log::severity_level::info);
+        using sl = nitro::log::severity_level;
+        switch (verbosity.count)
+        {
+        case 0:
+            lo2s::logging::set_min_severity_level(sl::warn);
```
tilsche: At least there should be a way to get `info`, but I'm not sure if it should be default.
How about:

. | .
------------ | -------------
`-q` | error
`(default)` | warn
`-v` | info
`-vv` | debug
`-vvv` | trace

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,389 @@
...
+ * along with lo2s.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <lo2s/config.hpp>
+#include <lo2s/log.hpp>
+#include <lo2s/perf/event_provider.hpp>
+
+#include <boost/filesystem.hpp>
+#include <regex>
+#include <sstream>
```
bmario: missing include <ios> to make IWYU happy

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,389 @@
...
+                 << std::dec << std::noshowbase;
+    return mask;
+}
+
+static constexpr std::uint64_t apply_mask(std::uint64_t value, std::uint64_t mask)
+{
+    std::uint64_t res = 0;
+    for (int mask_bit = 0, value_bit = 0; mask_bit < 64; mask_bit++)
+    {
+        if (mask & (1 << mask_bit))
```
bmario: (1 << mask_bit) also int

src/config.cpp
phijor: That sounds reasonable; I'll implement it according to @tilsche's suggestion.

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,407 @@
...
+        BIT_BEGIN,
+        BIT_END,
+    };
+
+    std::uint64_t mask = 0x0;
+
+    static const std::regex bit_mask_regex(R"((\d+)?(?:-(\d+)))");
+    const std::sregex_iterator end;
+    std::smatch bit_mask_match;
+    for (std::sregex_iterator i = { format.begin(), format.end(), bit_mask_regex }; i != end; i++)
```
bmario: Writing `i++` is a premature pessimization. Yes, the hope is that a sufficiently clever compiler could optimize the additional copy away, but looking at https://gcc.godbolt.org/#, gcc 7.1 doesn't. Generally, use prefix increment, unless you require the postfix.

phijor: Yeah, don't know why that's there. I use prefix increment everywhere else, do I not?

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,407 @@
...
+    }
+    Log::debug() << std::showbase << std::hex << "config mask: " << format << " = " << mask
+                 << std::dec << std::noshowbase;
+    return mask;
+}
+
+static constexpr std::uint64_t apply_mask(std::uint64_t value, std::uint64_t mask)
+{
+    std::uint64_t res = 0;
+    for (int mask_bit = 0, value_bit = 0; mask_bit < 64; mask_bit++)
```
bmario: Nope ;)

But in this particular case it doesn't matter, as it's an int.

src/perf/event_provider.cpp
```diff
@@ -0,0 +1,407 @@
...
+    {
+        if (mask & bit(mask_bit))
+        {
+            res |= ((value >> value_bit) & bit(0)) << mask_bit;
+            value_bit++;
```
bmario: Nope ;)

But in this particular case it doesn't matter, as it's an int.

bmario:
But it's okay. I will merge it, after Travis has done the things travis does. Whatever these things are.

PR #19, opened 2017-08-22
Issue 7, time synchronization:  add -k,--clockid options

phijor:


src/config.cpp
```diff
@@ -190,6 +193,50 @@ void parse_program_options(int argc, const char** argv)
...
+            },
+            {
+                "boottime", CLOCK_BOOTTIME,
+            },
+        };
+
+        const char* clock_name_ptr = clock_name.c_str();
+        for (auto clock : clocks)
+        {
+            if (std::strcmp(clock_name_ptr, clock.name) == 0)
```
bmario:
```
if (clock_name == clock.name)
```

src/config.cpp
```diff
@@ -190,6 +193,50 @@ void parse_program_options(int argc, const char** argv)
...
     }
 
+    // time synchronization
+    config.use_clockid = false;
+    if (!clock_name.empty())
+    {
+#if defined(USE_PERF_CLOCKID) && !defined(HW_BREAKPOINT_COMPAT)
+        struct clock_descripton
+        {
+            const char name[16];
```
bmario: Is there a reason, why this isn't a std::string?

src/config.cpp
```diff
@@ -190,6 +193,50 @@ void parse_program_options(int argc, const char** argv)
...
+            },
+            {
+                "realtime", CLOCK_REALTIME,
+            },
+            {
+                "boottime", CLOCK_BOOTTIME,
+            },
+        };
+
+        const char* clock_name_ptr = clock_name.c_str();
```
bmario: Not needed anymore after the next suggestion.

src/config.cpp
```diff
@@ -79,6 +79,8 @@ void parse_program_options(int argc, const char** argv)
     bool kernel, no_kernel;
     std::uint64_t read_interval_ms;
 
+    std::string clock_name;
```
bmario: Maybe rename to `requested_clock` or `requested_clock_name`

src/config.cpp
```diff
@@ -110,6 +112,7 @@ void parse_program_options(int argc, const char** argv)
              "enable global recording of a raw tracepoint event (usually requires root)")
         ("metric-event,E", po::value(&config.perf_events),
              "the name of a perf event to measure") // TODO: optionally list available events
+        ("clockid,k", po::value(&clock_name), "select timer to use for timestamps")
```
bmario: I'd change the description to something like:

"clock used for perf timestamps"

src/config.cpp
```diff
@@ -190,6 +193,50 @@ void parse_program_options(int argc, const char** argv)
...
+            {
+                "realtime", CLOCK_REALTIME,
+            },
+            {
+                "boottime", CLOCK_BOOTTIME,
+            },
+        };
+
+        const char* clock_name_ptr = clock_name.c_str();
+        for (auto clock : clocks)
```
bmario:
```
for (const auto& clock : clocks)
```

src/config.cpp
phijor: I wanted the struct to be usable as `constexpr`, that won't work with `std::string`s. But even then I'd think `const char*` might be better than `const char[16]` :thinking:

But if you think using `std::string` is more idiomatic I'll change it.

bmario: While `constexpr` certainly sounds like a valid point, it smells like premature optimization for me. 
`clock_description` is static anyways, so I don't see the benefit for `constexpr` here.
Using std::string is just less demanding in terms of brain computing capacity.

PR #20, opened 2017-09-05
Issue 7 time synchronization

bmario:


src/perf/time/converter.cpp
```diff
@@ -38,13 +38,27 @@ Converter::Converter()
...
-    offset = reader.local_time.time_since_epoch() - reader.perf_time.time_since_epoch();
-    Log::debug() << "perf time offset: " << offset.count() << " ns ("
-                 << reader.local_time.time_since_epoch().count() << " - "
-                 << reader.perf_time.time_since_epoch().count() << ").";
+
+    // we expect local_time <= perf_time, i.e. time_diff < 0
+    const auto time_diff =
+        reader.local_time.time_since_epoch() - reader.perf_time.time_since_epoch();
+
+    if (time_diff < std::chrono::microseconds(-100) or time_diff > std::chrono::microseconds(0))
```
bmario: This check should only be done if `lo2s::config().use_clockid` is true.

bmario: If this check fails, it might be a good idea to do this:

```
offset = time_diff;
```

include/lo2s/time/time.hpp
```diff
@@ -26,15 +26,99 @@
...
+            "boottime", CLOCK_BOOTTIME,
+        },
+        {
+            "realtime-alarm", CLOCK_REALTIME_ALARM,
+        },
+        {
+            "boottime-alarm", CLOCK_BOOTTIME_ALARM,
+        },
+        {
+            "tai", CLOCK_TAI,
```
phijor: Aaaaand there's systems where this isn't defined ಠ_ಠ (looking at you Travis).

We should probably only include clocks that are defined.

bmario: Why doesn't this help:

> The available timers are defined in linux/time.h

bmario: I remembered something: Travis uses very old linux header files, because Travis. And because it's terrible.
Therefore it doesn't even build with USE_PERF_CLOCKID=ON

So this might be the issue there.

phijor: So including `linux/time.h` still doesn't work (it clashes with `ctime`). But [POSIX](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/time.h.html) defines some of the `CLOCK_*` macros. I would really like a cleaner solution to this, but right now it seems like we have to rely on both headers defining the same values and continue to include from `ctime`.

bmario:
Fur future reference, we need to properly document the different clocks as well.
For instance the following note renders the two `ID` clocks unusable in most cases.

> Note for SMP systems
> 
> The CLOCK_PROCESS_CPUTIME_ID and CLOCK_THREAD_CPUTIME_ID clocks are realized on many platforms using timers from the CPUs (TSC on i386, AR.ITC on Itanium). These registers may differ between CPUs and as a consequence these clocks may return bogus results if a process is migrated to another CPU.

include/lo2s/time/time.hpp
```diff
@@ -26,15 +26,121 @@
...
+        }
+    };
+
+    static const ClockDescription& get_clock_by_name(const std::string& name);
+
+private:
+    static constexpr ClockDescription clocks_[] = {
+#ifdef CLOCK_REALTIME
+        {
+            "realtime", CLOCK_REALTIME,
```
bmario: Should be available on all implementations, no check needed.

include/lo2s/time/time.hpp
```diff
@@ -26,15 +26,121 @@
...
+        },
+#endif
+#ifdef CLOCK_MONOTONIC
+        {
+            "monotonic", CLOCK_MONOTONIC,
```
bmario: Check against _POSIX_MONOTONIC_CLOCK

include/lo2s/time/time.hpp
```diff
@@ -26,15 +26,121 @@
...
+        },
+#endif
+#ifdef CLOCK_PROCESS_CPUTIME_ID
+        {
+            "process-cputime-id", CLOCK_PROCESS_CPUTIME_ID,
```
bmario: Check against _POSIX_CPUTIME

include/lo2s/time/time.hpp
```diff
@@ -26,15 +26,121 @@
...
+        },
+#endif
+#ifdef CLOCK_THREAD_CPUTIME_ID
+        {
+            "process-thread-id", CLOCK_THREAD_CPUTIME_ID,
```
bmario: Check against _POSIX_THREAD_CPUTIME

include/lo2s/time/time.hpp
```diff
@@ -26,15 +26,121 @@
...
+        },
+#endif
+#ifdef CLOCK_BOOTTIME_ALARM
+        {
+            "boottime-alarm", CLOCK_BOOTTIME_ALARM,
+        },
+#endif
+#ifdef CLOCK_TAI
+        {
+            "tai", CLOCK_TAI,
```
bmario: I think TAI is not necessary, as it's equivalent to REALTIME

include/lo2s/time/time.hpp
```diff
@@ -26,15 +26,121 @@
...
+        },
+#endif
+#ifdef CLOCK_THREAD_CPUTIME_ID
+        {
+            "process-thread-id", CLOCK_THREAD_CPUTIME_ID,
+        },
+#endif
+#ifdef CLOCK_MONOTONIC_RAW
+        {
+            "monotonic-raw", CLOCK_MONOTONIC_RAW,
```
bmario: Check other CLOCKIDs with corresponding Linux version checks.

See:
http://man7.org/linux/man-pages/man2/timerfd_create.2.html
https://linux.die.net/man/2/clock_gettime

PR #21, opened 2017-10-02
Calculate the correct mmap-page value on the fly

cvonelm:
uses
(/proc/sys/kernel/perf_event_mlock_kb * 1000 )/ pagesize -1
to calculate the correct value

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -68,7 +68,30 @@ const Config& config()
 {
     return *instance;
 }
+int get_default_mmap_value()
+{
+    int retval = 16;
+    std::fstream mlock_kb_file;
+    //max mlock()able pages per user (in kilobytes)
+    int mlock_kb;
+    //We are only interested in the calculated value,
+    //so a catch-all approach works
+    try 
+    {
+        mlock_kb_file.open(
+                "/proc/sys/kernel/perf_event_mlock_kb",
+                std::fstream::in);
+        mlock_kb_file >> mlock_kb;
+        mlock_kb_file.close();
 
+        long pagesize = sysconf(_SC_PAGESIZE);
+        
+        if(pagesize > 0)
+            retval = (mlock_kb * 1024) / pagesize - 1;
+    }
+    catch (...){}
+    return retval;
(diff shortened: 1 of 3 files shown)


cvonelm:
Sorry the formula is
/proc/sys/kernel/perf_event_mlock_kb  * 1024)/pagesize -1 instead, assuming 1024 bytes in a kilobyte

src/config.cpp
```diff
@@ -68,7 +68,30 @@ const Config& config()
...
+        mlock_kb_file.open(
+                "/proc/sys/kernel/perf_event_mlock_kb",
+                std::fstream::in);
+        mlock_kb_file >> mlock_kb;
+        mlock_kb_file.close();
 
+        long pagesize = sysconf(_SC_PAGESIZE);
+        
+        if(pagesize > 0)
+            retval = (mlock_kb * 1024) / pagesize - 1;
```
bmario: Please add the curly braces even for that one line if

src/config.cpp
```diff
@@ -68,7 +71,32 @@ const Config& config()
 {
     return *instance;
 }
+int get_default_mmap_value()
```
tilsche: Please move the function to `util.cpp` with all the other `/proc` / `sysconf` wrappers. I would suggest to name it there `get_perf_event_mlock` and return a `size_t` value in bytes, or in case of an exception a default of 516*1024.

src/config.cpp
```diff
@@ -107,7 +134,7 @@ void parse_program_options(int argc, const char** argv)
              "suppress output")
         ("verbose,v", po::value(&verbosity)->zero_tokens(),
              "verbose output (specify multiple times to get increasingly more verbose output)")
-        ("mmap-pages,m", po::value(&config.mmap_pages)->default_value(16),
+        ("mmap-pages,m", po::value(&config.mmap_pages)->default_value(get_default_mmap_value()),
```
tilsche: In accordance with the previous change, use `get_perf_event_mlock() - get_page_size()` as default.

PR #26, opened 2017-10-31
Issue 8: group metrics and read them from a mmap buffer

phijor:
...and other assorted fixes.

Commit 56ff0dc mostly explains what this is about. Counters are now grouped together (see 75f07e4), and their values aren't read from a file descriptor anymore but instead are periodically stored in the mmap ringbuffer of the leader. They are later read by `perf::counter::Writer` and written to `otf2`.


include/lo2s/perf/counter/reader.hpp
```diff
@@ -0,0 +1,81 @@
...
+    {
+        double read_interval = config().read_interval.count();
+        Log::debug() << "perf::counter::Reader: sample_freq: " << 1.0e9 / read_interval << "Hz";
+
+        static struct perf_event_attr leader_attr;
+        std::memset(&leader_attr, 0, sizeof(leader_attr));
+
+        leader_attr.size = sizeof(leader_attr);
+        leader_attr.freq = 1;
+        leader_attr.sample_freq = static_cast<uint64_t>(1.0e9 / read_interval);
```
tilsche: I would like to at least support using a fixed event period rather than the perf frequency thingy:

> The kernel will adjust the sampling period to try and achieve the desired rate.  The rate of adjustment is a timer tick.

That sounds like it may have varying results.

The interesting question is how to allow the user to properly specify group leader / periods and possible for easier access frequencies. We can keep it like proposed for now, but I'd like to keep that for further improvement.

src/perf/event_provider.cpp
```diff
@@ -145,7 +146,7 @@ static bool event_is_openable(const CounterDescription& ev)
     attr.type = ev.type;
     attr.config = ev.config;
     attr.config1 = ev.config1;
-    attr.exclude_kernel = 1;
+    attr.exclude_kernel = 0;
 
     int fd = syscall(__NR_perf_event_open, &attr, 0, -1, -1, 0);
```
tilsche: use the new util wrapper?

src/metric/perf_counter.cpp
```diff
@@ -85,5 +99,109 @@ double PerfCounter::read()
...
+    }
+
+    update_buffers();
+}
+
+void CounterBuffer::read(const ReadFormat* inbuf)
+{
+    assert(accumulated_.size() == inbuf->nr);
+
+    auto ret = std::memcpy(current_, inbuf, total_buf_size(inbuf->nr));
```
tilsche: As per the standard `std::memcpy` always returns `dest` - there's no use in checking. When it fails, it segfaults.

src/config.cpp
```diff
@@ -81,6 +82,7 @@ void parse_program_options(int argc, const char** argv)
     bool kernel, no_kernel;
     bool list_clockids, list_events;
     std::uint64_t read_interval_ms;
+    boost::optional<std::uint64_t> metric_count, metric_frequency;
```
phijor: So apparently, reading into `boost::optional<T>` is only supported since boost 1.65. I will think of another way to parse these mutually exclusive flags and at the same time give them sane defaults.

bmario:
There is an actual build issue on travis.

phijor:
Yeah, that is actually my fault. Reading program options into `boost::optional`s was only introduced with boost 1.65. See [here](https://github.com/tud-zih-energy/lo2s/pull/26#discussion_r152403869).

src/config.cpp
tilsche: Usually I'm the first one to update the dependency version, but there may be some concern with availability on wide spread distributions. boost 1.65 appears to be 3 months old. That said, Ubuntu LTS only comes with the ancient boost 1.58 which we don't currently support anyway... Ubuntu 17.* has 1.62.0. @bmario, @rschoene what do you think about bumping the boost version requirements?

bmario: I'm against bumping boost version for such a minor issue. Just properly handle it in `lo2s::config::parse_program_options()`. There is `vm.count()`.

phijor: I too think that this issue should *not* warrant updating boost.

The only caveat of `vm.count()` is that if `po::value::default_value()` is used on an option, `vm.count()` is always non-zero, even if that option was not specified. One could argue that this is the intended behavior, but right now I don't see how we would display a default value to the user that we ignore if the option was not specified.

<sup>But using `boost::optional` doesn't solve that issue either, I just realised</sup>

tilsche:
Pending performance test, three scenarios

- benchmark, normal execution
- benchmark, lo2s master
- benchmark, lo2s branch
- perf record with similar configuration

Use default set of metrics.
Configure readout interval such that there is noticeable overhead for both lo2s cases.

phijor:
After finally getting all tests set up and making `lo2s` read samples again, here are some measurements on what happens with the new reading mechanism compared to the old implementation. Test case was the NPB BT benchmark, class C. To increase the impact `lo2s` has on execution, the metric readout interval was set to 1ms (the minimum possible on `master`).

| instrumentation                                     | Time in seconds | Mop/s total | Mop/s/thread |
|:----------------------------------------------------|----------------:|------------:|-------------:|
| none                                                |           68.54 |    41818.98 |      1306.84 |
| `lo2s` (branch `master`)                            |           79.02 |    36273.93 |      1133.56 |
| `lo2s` (branch `issue-8-configurable-perf-metrics`) |           74.44 |    38502.12 |      1203.19 |

Should I post any other measurements/comparisons? I've got a bunch of otf2 traces that might be interesting for comparison.

Bonus picture (because it looks nice :D):

`lo2s` measuring the benchmark with default metrics/readout intervals. Top is `master` using the interval timer, bottom `issue-8-configurable-perf-metrics` using the `mmap`-buffer. Displayed metric is `time_enabled`.

lo2s_comparison_master_issue-8

tilsche:
Sounds great! Can you post the command line used for the test with `issue-8-configurable-perf-metrics`? Did you visually confirm the metric interval for this case to be around 1 ms?

phijor:
I tested with
```
$ ./lo2s --metric-frequency 1000 -- bt.C.x
```
and some output redirection (so traces would be written to `/fastfs`). Metric interval is almost always very close to 1ms. The only exception is when a thread unexpectedly idles: It can take a couple ms for it to generate another sample then.

Here's a rather extreme example (compare thread `4611:1`):

idle

I think this is something we could prevent from happening by using a fallback timer (as @rschoene suggested) or maybe finding a reference counter that steadily increases even when the thread is idle.

tilsche:
Can you try with `ref-cycles` and a period that will be equivalent to 1 ms?

phijor:
I put a system on a nominal frequency of 2.9GHz and running
```
$ ./lo2s --metric-count 2900000 --metric-leader ref-cycles -- bt.C.x
```
(consistently) yielded the expected readout intervals of ~1ms.

PR #36, opened 2018-01-22
Issue 31 measurement summary

cvonelm:
This pull request adds a short measurement summary looking like this:

process mode:

```bash
[ lo2s: bash  (0), 3 threads, 0.09s CPU, 6.5s total ]
[ lo2s: 148 wakeups, wrote 10.9 KiB lo2s_trace_2018-01-22T13-18-49 ]
```
-a mode:

```bash
[ lo2s: -a mode, sampled processes: 62, 0.05s CPU, 3.64s total ]
[ lo2s: 4 wakeups, wrote 58.18 KiB lo2s_trace_2018-01-22T13-19-46 ]
```
To confirm that i havent missed a spot where threads are created, processes are sampled etc. here are
how these numbers get calculated

|Metric | How its calculated |
|---------|--------------------------|
threads  | increased for every event that adds a new thread to the thread map in ProcessMonitor
wakeups | increased for every time ProcessMonitor.handle_signal(), IntervalMonitor.monitor() and FdMonitor.monitor gets called from the classes main loop
sampled processes | increased for every new thread we encounter in ExitReader and SwitchWriter


bmario:
The summary looks good so far, but one suggestion:

In -a mode:

```
[ lo2s: 62 monitored processes, 0.05s CPU, 3.64s total ]
[ lo2s: 4 wakeups, wrote 58.18 KiB lo2s_trace_2018-01-22T13-19-46 ]
```

phijor:
One suggestion:

Instead of doing
```
Summary::increase_thread_count();
threads_.insert(child, child, spawn);
```
everywhere, why not make `lo2s::ThreadMap::insert()` increase the global thread count?

And a question:
Is `lo2s::Summary::thread_count_` accessed atomically? Or is it only written to by a single thread?

include/lo2s/summary.hpp
```diff
@@ -0,0 +1,26 @@
...
+{
+class Summary
+{
+public:
+    static void start();
+    static void finalize_and_print();
+    static void record_wakeups(int num_wakeups);
+    static void increase_thread_count();
+    static void set_trace_dir(std::string trace_dir);
+    static void set_pid(pid_t pid);
```
bmario: I'd rename it to a more high-level descriptive name like `register_process` or something.

include/lo2s/summary.hpp
```diff
@@ -0,0 +1,26 @@
...
+namespace lo2s
+{
+class Summary
+{
+public:
+    static void start();
+    static void finalize_and_print();
+    static void record_wakeups(int num_wakeups);
+    static void increase_thread_count();
+    static void set_trace_dir(std::string trace_dir);
```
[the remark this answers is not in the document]
cvonelm: The trace_path member of config is just the prefix of the trace_path, the date part gets added later in the constructor of Trace in trace/trace.cpp by get_trace_name().We could move that function to config and have config().trace_path name the full path as get_trace_name() has no apparent ties to anything in trace/trace.cpp

include/lo2s/summary.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <ctime>
+namespace lo2s
+{
+class Summary
+{
+public:
+    static void start();
+    static void finalize_and_print();
+    static void record_wakeups(int num_wakeups);
+    static void increase_thread_count();
```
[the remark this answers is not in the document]
cvonelm: Im unsure about how much PID reuse would affect us there, so I left it in for now

CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ if(lo2s_USE_STATIC_LIBS STREQUAL "ALL")
 endif()
 
 
+
```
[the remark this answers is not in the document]
cvonelm: Sorry, small oversight from hand merging CMakeLists.txt after we updated the cmake version

src/summary.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+    exit_code_ = exit_code;
+}
+
+void Summary::add_thread()
+{
+    thread_count_++;
+}
+
+std::string pretty_print_bytes(int trace_size)
```
bmario: please use std::size_t trace_size. int has only 32-bit, thus would only allow for 2GB.

src/summary.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+    double result_size = trace_size;
+    int unit = 0;
+    std::string units[5] = { "B", "KiB", "MiB", "GiB", "TiB" };
```
bmario: please use std::array<>, so you can replace the magic "4" a few lines below with `units.size()`

src/summary.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    for (; result_size > 1024; unit++, result_size /= 1024)
```
bmario: I don't like this usage of `for`, as it's not in the canonical form and thus harder to parse for a human. Actually, I'd prefer a while loop here.

src/summary.cpp
```diff
@@ -0,0 +1,124 @@
...
+        ;
+
+    if (unit > 4)
```
bmario: It would be better to print a bigger number of bytes in TiB instead. You could simply add an additional condition to the proposed while loop.

src/summary.cpp
```diff
@@ -49,54 +50,71 @@ void Summary::add_thread()
...
-std::string pretty_print_bytes(int trace_size)
+std::string pretty_print_bytes(std::size_t trace_size)
 {
     double result_size = trace_size;
     int unit = 0;
-    std::string units[5] = { "B", "KiB", "MiB", "GiB", "TiB" };
 
-    for (; result_size > 1024; unit++, result_size /= 1024)
-        ;
+    std::ostringstream out;
```
bmario: the variable was declared at the right spot before this change. Always declare variables as late as possible and preferable only, when you can initialize them.

src/summary.cpp
```diff
@@ -0,0 +1,140 @@
...
+    std::array<std::string, 5> units = { "B", "KiB", "MiB", "GiB", "TiB" };
+
+    while(result_size > 1024)
+    {
+        result_size /= 1024;
+        unit++;
+
+        //We can not get higher than TiB so break here
+        if(unit == units.size() - 1)
```
bmario: This is a comparison between signed and unsigned. I don't know why there is no compiler warning, but an easy fix would be:
unit should be std::size_t

src/summary.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+void Summary::set_trace_dir(const std::string& trace_dir)
+{
+    trace_dir_ = trace_dir;
+}
+
+void Summary::show()
+{
+    std::chrono::duration<double> wall_time, cpu_time;
```
bmario: Same as above, please place variable declarations as close as possible to the first use. Also, never ever declare two variables on one line.

src/summary.cpp
```diff
@@ -0,0 +1,140 @@
...
+        {
+            std::cout << command << ' ';
+        }
+
+        std::cout << " (" << exit_code_ << "), ";
+        std::cout << thread_count_ << " threads, ";
+    }
+    else
+    {
+        std::cout << "-a mode, ";
```
bmario: Should we change `-a mode` to `system mode` or even `system monitoring`?
@tilsche

src/summary.cpp
```diff
@@ -0,0 +1,140 @@
...
+        unit++;
+
+        //We can not get higher than TiB so break here
+        if(unit == units.size() - 1)
+        {
+            break;
+        }
+    }
+
+    out << std::fixed << std::setprecision(2) << result_size << units[unit];
```
bmario: A space character is missing between the number and the unit.

include/lo2s/summary.hpp
```diff
@@ -0,0 +1,45 @@
...
+private:
+    Summary();
+
+    std::chrono::steady_clock::time_point start_wall_time_;
+
+    std::atomic<std::size_t> num_wakeups_;
+    std::atomic<std::size_t> thread_count_;
+
+    std::unordered_set<pid_t> pids_;
+    std::mutex pids_mutex_;
```
bmario: This mutex isn't used anywhere though it should.

src/summary.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+Summary::Summary()
+: start_wall_time_(std::chrono::steady_clock::now()), num_wakeups_(0),
+  thread_count_(0), trace_dir_(""), exit_code_(0)
+{
+}
+
+void Summary::register_process(pid_t pid)
+{
+    pids_.emplace(pid);
```
bmario: This should use a lock guard on pids_mutex_

src/summary.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+Summary& summary()
+{
+    static Summary s;
+    return s;
+}
+
+Summary::Summary()
+: start_wall_time_(std::chrono::steady_clock::now()), num_wakeups_(0),
+  thread_count_(0), trace_dir_(""), exit_code_(0)
```
bmario: `trace_dir_("")` is equivalent to the default constructor of std::string, so you can omit it. Less code is better code.

src/summary.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    std::size_t trace_size;
+
+    if (config().quiet)
+    {
+        return;
+    }
+
+    std::chrono::duration<double> wall_time;
+    wall_time = std::chrono::steady_clock::now() - start_wall_time_;
```
bmario: Use one statement for definition and initialization.

src/summary.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    std::chrono::duration<double> cpu_time;
+    cpu_time = get_cpu_time();
```
bmario: Use one statement for definition and initialization.

Issue #39, opened 2018-01-31
Add lo2s version


tilsche:
* cmake gets git commit hash or tag
* add output on help page // `--version`
* add info in creator otf2 archive metadata

PR #42, opened 2018-02-01
Issue 40 Add full lo2s arguments to trace archive metadata

cvonelm:
This pull requests adds code to set the creator field of the otf2 archive metadata to the used command line.

### DIFF
--- a/src/monitor/main_monitor.cpp
+++ b/src/monitor/main_monitor.cpp
@@ -41,6 +41,8 @@ MainMonitor::MainMonitor()
     // the first possible timestamp in the trace
     trace_.begin_record();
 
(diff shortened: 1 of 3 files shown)


src/monitor/main_monitor.cpp
```diff
@@ -41,6 +41,8 @@ MainMonitor::MainMonitor()
     // the first possible timestamp in the trace
     trace_.begin_record();
 
+    trace_.archive().set_creator(config().command_line);
```
bmario: The creator should be lo2s with some version information. See #39. IMHO `set_description()` fits better here. If it does not exist in otf2xx yet, then make a PR there as well, please.

Issue #32, opened 2017-12-15
add sampling source as counter


rschoene:
When specifying a different sampling event (e.g., via `-e cpu-cycles`), the event is not recorded as a counter. Please add it.

phijor:
I would like to address this as soon as PR #26 is merged. It has the infrastructure in place to collect counters in a central place, which makes this an easy change.

rschoene:
I would have preferred (1). Since this is "problematic", it should not be implemented.

PR #46, opened 2018-02-15
Add sampling source event as a counter

phijor:
If not already requested by `-E` or `--metric-leader`, the sampling source event specified by `-e` is added as a metric counter.

This should implement the behavior requested in issue #32.


bmario:
As discussed in #32, this doesn't go in the right direction. So I'm closing it for now.

Issue #49, opened 2018-02-16
Sort and group options in usage message


bmario:
This is possible when using more than one `po::options_description`.

The usage page should look more like this:

```
Usage:
  ./lo2s [options] ./a.out
  ./lo2s [options] -- ./a.out --option-to-a-out
  ./lo2s [options] --pid $(pidof some-process)

Allowed options:
  --help                                produce help message
  --version                             print version information
  -q [ --quiet ]                        suppress output
  -v [ --verbose ]                      verbose output (specify multiple times 
                                        to get increasingly more verbose 
                                        output)
  -o [ --output-trace ] arg             output trace directory
  --list-clockids                       list all available clockids
  --list-events                         list all available events
  -m [ --mmap-pages ] arg (=16)         number of pages to be used by each 
                                        internal buffer
  -k [ --clockid ] arg (=monotonic-raw) clock used for perf timestamps (see 
                                        --list-clockids for supported 
                                        arguments)

System-wide monitoring:
  -a [ --all-cpus ]                     System-wide monitoring of all CPUs.

Sampling options:
  --command arg
  -c [ --count ] arg (=11010113)        sampling period (# of events specified 
                                        by -e)
  -e [ --event ] arg (=instructions)    interrupt source event for sampling
  -g [ --call-graph ]                   call-graph recording
  -n [ --no-ip ]                        do not record instruction pointers [NOT
                                        CURRENTLY SUPPORTED]
  -p [ --pid ] arg (=-1)                attach to specific pid
  -i [ --readout-interval ] arg (=100)  time interval between metric and 
                                        sampling buffer readouts in 
                                        milliseconds
  --disassemble                         enable augmentation of samples with 
                                        instructions (default if supported)
  --no-disassemble                      disable augmentation of samples with 
                                        instructions
  --kernel                              include events happening in kernel 
                                        space (default)
  --no-kernel                           exclude events happening in kernel 
                                        space

Kernel trace point options:
  -t [ --tracepoint ] arg               enable global recording of a raw 
                                        tracepoint event (usually requires 
                                        root)

Perf metric options:
  -E [ --metric-event ] arg             the name of a perf event to measure
  --metric-leader arg (=ref-cycles)     name of leading perf event
  --metric-count arg                    # of events to elapse by metric leader 
                                        before reading metric buffer
  --metric-frequency arg                metric buffer reads per second

x86_adapt options:
[...]

PR #50, opened 2018-03-09
Issue #49: usability enhancements

phijor:
Changes:

* list available events (predefined and kernel PMU) with `--list-events`
* list available tracepoint events with `--list-tracepoints`
* show descriptive error when using `x86_adapt`-related switches but `lo2s` was built without support for `x86_adapt`.
* group options in usage section
* add `lo2s [OPTIONS] -a` as invocation pattern in usage section
* give names to arguments to options
* add usage section explaining arguments to options.

### DIFF
--- a/include/lo2s/io.hpp
+++ b/include/lo2s/io.hpp
@@ -1,0 +1,124 @@
+/*
+ * This file is part of the lo2s software.
+ * Linux OTF2 sampling
+ *
+ * Copyright (c) 2018,
+ *    Technische Universitaet Dresden, Germany
+ *
+ * lo2s is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * lo2s is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with lo2s.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * \file lo2s/io.hpp
+ * \brief I/O helper functions for lo2s' types
+ *
+ * This header contains overloads for stream-based formatted output of types
+ * used by lo2s and I/O-manipulators for uniform looking output.
+ *
+ */
+
+#pragma once
+
+#include <iostream>
+
+#include <lo2s/time/time.hpp>
+
+namespace lo2s
+{
+
+namespace time
+{
+inline std::ostream& operator<<(std::ostream& os, const ClockDescription& desc)
+{
+    return os << desc.name;
+}
+}
+
+namespace io
+{
+
+/**
+ * \brief Format a range of values as a list of arguments to an option.
+ */
+template <class InputIterator>
+struct ArgumentList
+{
+    ArgumentList() = delete;
+
+    /**
+     * \param description   description of list contents printed before list
+     * \param first         iterator pointing to first item to print
+     * \param last          iterator pointing past the last item to print
+     * */
+    constexpr ArgumentList(const char* description, InputIterator first, InputIterator last)
+    : description_(description), first_(first), last_(last)
+    {
+    }
+
+    template <class _InputIt>
+    friend ArgumentList<InputIterator> make_argument_list(const char*, _InputIt, _InputIt);
+
+    template <class T, std::size_t N>
+    friend ArgumentList<InputIterator> make_argument_list(const char* description,
+                                                          const T (&array)[N]);
+
+    template <class _InputIt>
+    friend std::ostream& operator<<(std::ostream&, const ArgumentList<_InputIt>&);
+
+private:
+    const char* description_;
+    InputIterator first_, last_;
+};
+
+/**
+ * \brief   Creates ArgumentList object, deducing the target type from the types
+ *          of the arguments
+ */
+template <class InputIterator>
+inline constexpr auto make_argument_list(const char* description, InputIterator first,
+                                         InputIterator last)
+{
+    return ArgumentList<InputIterator>{ description, first, last };
+}
+
+/**
+ * \brief Creates ArgumentList object from a fixed-size array of elements
+ */
+template <class T, std::size_t N>
+inline constexpr auto make_argument_list(const char* description, const T (&array)[N])
+{
+    return ArgumentList<const T*>{ description, array, array + N };
+}
+
+template <class InputIterator>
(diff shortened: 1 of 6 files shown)


bmario:
Looks good, but I'd still like to improve a few things:

- [ ] `--list-events=tracepoint` should actually be `--list-tracepoints`
- [ ] There should also be a  `--list-knobs`
- [ ] the argument for `-x [ --x86-adapt-cpu-knob ] arg` should be KNOB
- [ ] `--command` shouldn't be listed as option

src/config.cpp
```diff
@@ -103,99 +189,170 @@ const Config& config()
...
+                ->default_value("monotonic-raw"),
+            "Clock used for perf timestamps (see --list-clockids for supported arguments).")
+        ("list-clockids",
+            po::bool_switch(&list_clockids)
+                ->default_value(false),
+            "List all available clockids.")
+        ("list-events",
+            po::value(&list_event_category)
+                ->value_name("CATEGORY")
+                ->implicit_value("all"),
```
bmario: The help message generated from this looks different than other options.
Considering that I'd like to split this anyways, just turn this into a bool_switch and print predefined and pmu events.
`--list-tracepoints` than should print the list of tracepoints.

src/config.cpp
```diff
@@ -103,99 +189,170 @@ const Config& config()
...
+            po::bool_switch(&config.enable_cct),
+            "Enable call-graph recording.")
+        ("no-ip,n",
+            po::bool_switch(&config.suppress_ip),
+            "Do not record instruction pointers [NOT CURRENTLY SUPPORTED]")
+        ("pid,p",
+            po::value(&config.pid)
+                ->value_name("PID")
+                ->default_value(-1),
+            "Attach to process of given PID.")
```
bmario: Can we hide the `(=-1)` in the output? Probably something like remove the `default_value()` and add something like this: 

```
if vm.count("pid"):
  config.pid = vm.pid
else:
  config.pid = -1
```

src/config.cpp
```diff
@@ -103,99 +189,170 @@ const Config& config()
...
+                ->default_value(11010113),
+            "Sampling period (in number of events specified by -e).")
+        ("event,e",
+            po::value(&config.sampling_event)
+                ->value_name("EVENT")
+                ->default_value("instructions"),
+            "Interrupt source event for sampling.")
+        ("call-graph,g",
+            po::bool_switch(&config.enable_cct),
+            "Enable call-graph recording.")
```
bmario: Change help to:
`Record call stack of samples`

src/config.cpp
```diff
@@ -86,12 +148,36 @@ static inline void print_version(std::ostream& os)
...
+                usually requires read/execute permissions on
+                    /sys/kernel/debug,
+                which by default is owned by root:root.
+)";
+
     // clang-format off
     os << "Usage:\n"
           "  " << name << " [options] ./a.out\n"
           "  " << name << " [options] -- ./a.out --option-to-a-out\n"
           "  " << name << " [options] --pid $(pidof some-process)\n"
```
bmario: Please also add a line for `-a` mode

src/config.cpp
```diff
@@ -86,12 +128,39 @@ static inline void print_version(std::ostream& os)
...
+    TRACEPOINT  Name of a kernel tracepoint event.  Format is either
+                    <group>:<name>
+                or
+                    <group>/<name>.
+                Tracepoint events can be found at
+                    /sys/kernel/debug/tracing/events/<group>/<name>.
+                Accessing tracepoints (even for --list-tracepoints)
+                usually requires read/execute permissions on
+                    /sys/kernel/debug
+)";
```
bmario: This should also contain an entry for `KNOB` and `CLOCKID`.
Also, mention the list command for each one.

src/config.cpp
```diff
@@ -240,25 +393,35 @@ void parse_program_options(int argc, const char** argv)
...
-    if (list_clockids || list_events)
     {
         if (list_clockids)
         {
-            std::cout << "Available clockids:\n";
-            for (const auto& clock : lo2s::time::ClockProvider::get_descriptions())
-            {
-                std::cout << " * " << clock.name << '\n';
-            }
+            std::cout << io::make_argument_list("available clockids",
```
bmario: This should more look like:

```
auto clock_ids = time::ClockProvider::get_descriptions();
std::cout << io::make_argument_list("available clockids", std::begin(clock_ids), std::end(clock_ids));
```

include/lo2s/io.hpp
```diff
@@ -0,0 +1,124 @@
...
+                                         InputIterator last)
+{
+    return ArgumentList<InputIterator>{ description, first, last };
+}
+
+/**
+ * \brief Creates ArgumentList object from a fixed-size array of elements
+ */
+template <class T, std::size_t N>
+inline constexpr auto make_argument_list(const char* description, const T (&array)[N])
```
bmario: This overload is not necessary, use `std::begin()` and `std::end()` instead. See below.

include/lo2s/io.hpp
```diff
@@ -0,0 +1,124 @@
...
+struct ArgumentList
+{
+    ArgumentList() = delete;
+
+    /**
+     * \param description   description of list contents printed before list
+     * \param first         iterator pointing to first item to print
+     * \param last          iterator pointing past the last item to print
+     * */
+    constexpr ArgumentList(const char* description, InputIterator first, InputIterator last)
```
bmario: > `const char*`

We are not in C, please use `const std::string&` instead.

src/config.cpp
```diff
@@ -73,6 +80,41 @@ void validate(boost::any& v, const std::vector<std::string>&, SwitchCounter*, lo
     }
 }
 
+static inline void list_arguments_sorted(std::ostream& os, const char* description,
```
bmario: const std::string&

src/perf/tracepoint/format.cpp
```diff
@@ -98,6 +98,29 @@ void EventFormat::parse_format_line(const std::string& line)
     }
 }
 
+std::vector<std::string> EventFormat::get_tracepoint_event_names() try
```
bmario: You're missing a pair of curly braces.

src/config.cpp
```diff
@@ -86,12 +128,39 @@ static inline void print_version(std::ostream& os)
 static inline void print_usage(std::ostream& os, const char* name,
                                const po::options_description& desc)
 {
+    static constexpr char argument_detail[] =
+        R"(
+Arguments to options:
+    PATH        A filesystem path.  It can be either relative to the
```
bmario: Please include the changes of ae40d227de88d221d3151bf63ed218c959d04f4f on master

src/perf/tracepoint/format.cpp
phijor: Today I learned that this is actually a valid syntax construct: a [function-try-block](http://en.cppreference.com/w/cpp/language/function-try-block). Will change nonetheless.

PR #51, opened 2018-03-09
Issue #5: process tree

phijor:
Add traced processes in a tree to the trace.

As taken from 6422a2c:

```
Suppose on machine HAL 9000, lo2s traced a process "foo" which spawned
some threads and another process "bar", the resulting system tree now
looks something like this:

<root node>
`-- <machine "HAL 9000">
    `-- <process "foo">         \
    |   `-- <thread 1234>       |
    |   `-- ...                 |
    |   `-- <thread 1337>       | new process subtree
    |   `-- <process "bar">     |
    |       `-- ...             /
    `-- <Metric Location Group>
    `-- <package 0>
    `-- ...
    `-- <package n>
```

### DIFF
--- a/src/trace/trace.cpp
+++ b/src/trace/trace.cpp
@@ -80,7 +80,6 @@ Trace::Trace()
     // TODO clean this up, avoid side effect comm stuff
     process(METRIC_PID, "Metric Location Group");
 
-    int otf2_id = 1;
     const auto& sys = Topology::instance();
     for (auto& package : sys.packages())
     {
@@ -88,8 +87,8 @@ Trace::Trace()
         const auto& parent = system_tree_root_node_;
         system_tree_package_nodes_.emplace(
             std::piecewise_construct, std::forward_as_tuple(package.id),
-            std::forward_as_tuple(otf2_id++, intern(std::to_string(package.id)), intern("package"),
-                                  parent));
+            std::forward_as_tuple(system_tree_ref(), intern(std::to_string(package.id)),
+                                  intern("package"), parent));
     }
     for (auto& core : sys.cores())
     {
@@ -98,7 +97,8 @@ Trace::Trace()
         system_tree_core_nodes_.emplace(
             std::piecewise_construct, std::forward_as_tuple(core.id, core.package_id),
             std::forward_as_tuple(
-                otf2_id++, intern(std::to_string(core.package_id) + ":"s + std::to_string(core.id)),
+                system_tree_ref(),
+                intern(std::to_string(core.package_id) + ":"s + std::to_string(core.id)),
                 intern("core"), parent));
     }
     for (auto& cpu : sys.cpus())
@@ -107,7 +107,7 @@ Trace::Trace()
         const auto& parent =
             system_tree_core_nodes_.at(std::make_pair(cpu.core_id, cpu.package_id));
         system_tree_cpu_nodes_.emplace(std::piecewise_construct, std::forward_as_tuple(cpu.id),
-                                       std::forward_as_tuple(otf2_id++,
+                                       std::forward_as_tuple(system_tree_ref(),
                                                              intern(std::to_string(cpu.id)),
                                                              intern("cpu"), parent));
     }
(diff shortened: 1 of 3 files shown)


src/trace/trace.cpp
```diff
@@ -286,7 +339,7 @@ otf2::definition::metric_member Trace::metric_member(const std::string& name,
                                                      otf2::common::type value_type,
                                                      const std::string& unit)
 {
-    auto ref = metric_members_.size();
+    auto ref = metric_member_ref();
```
bmario: with these `*_ref()` methods now in place, I'd rather remove this local variable.

src/trace/trace.cpp
```diff
@@ -296,7 +349,7 @@ otf2::definition::metric_instance
 Trace::metric_instance(otf2::definition::metric_class metric_class,
                        otf2::definition::location recorder, otf2::definition::location scope)
 {
-    auto ref = metric_instances_.size() + metric_classes_.size();
+    auto ref = metric_instance_ref();
```
bmario: with these *_ref() methods now in place, I'd rather remove this local variable.

src/trace/trace.cpp
```diff
@@ -305,13 +358,13 @@ Trace::metric_instance(otf2::definition::metric_class metric_class,
                        otf2::definition::location recorder,
                        otf2::definition::system_tree_node scope)
 {
-    auto ref = metric_instances_.size() + metric_classes_.size();
+    auto ref = metric_instance_ref();
```
bmario: with these *_ref() methods now in place, I'd rather remove this local variable.

src/trace/trace.cpp
```diff
@@ -305,13 +358,13 @@ Trace::metric_instance(otf2::definition::metric_class metric_class,
...
     return metric_instances_.emplace(ref, metric_class, recorder, scope);
 }
 
 otf2::definition::metric_class Trace::metric_class()
 {
-    auto ref = metric_instances_.size() + metric_classes_.size();
+    auto ref = metric_class_ref();
```
bmario: with these *_ref() methods now in place, I'd rather remove this local variable.

src/trace/trace.cpp
```diff
@@ -509,7 +562,7 @@ otf2::definition::comm Trace::process_comm(pid_t pid)
 
 otf2::definition::source_code_location Trace::intern_scl(const LineInfo& info)
 {
-    auto ref = source_code_locations_.size();
+    auto ref = scl_ref();
```
bmario: with these *_ref() methods now in place, I'd rather remove this local variable.

src/trace/trace.cpp
```diff
@@ -530,7 +583,7 @@ otf2::definition::region Trace::intern_region(const LineInfo& info)
 
 otf2::definition::string Trace::intern(const std::string& name)
 {
-    auto ref = strings_.size();
+    auto ref = string_ref();
```
bmario: with these *_ref() methods now in place, I'd rather remove this local variable.

src/trace/trace.cpp
```diff
@@ -530,7 +583,7 @@ otf2::definition::region Trace::intern_region(const LineInfo& info)
...
     auto ret = strings_.emplace(std::piecewise_construct, std::forward_as_tuple(name),
```
bmario: To avoid costly constructions of `definition::string` objects, this should be a `try_emplace`.

src/trace/trace.cpp
```diff
@@ -209,40 +211,91 @@ std::map<std::string, otf2::definition::regions_group> Trace::regions_groups_sam
...
+
+void Trace::process_update_executable(pid_t pid, const otf2::definition::string& exe_name)
+{
+    system_tree_process_nodes_.at(pid).name(exe_name);
+    location_groups_process_.at(pid).name(exe_name);
+
+    // TODO: add setters to otf2xx to also update the names in comms and comm
+    // groups
+
+    // process_comm_groups_.at(pid).name(exe_name);
```
bmario: Please make this PR against otf2xx before merging this.

Issue #41, opened 2018-01-31
Handle cmd argument on lo2s -a


tilsche:
If a command is passed `lo2s -a -- ./command`. The command is spawned and lo2s will record until it is finished - but not monitor it otherwise.

cvonelm:
I now have tried to implement this using the current ProcessMonitor, disabling the code paths specific to monitoring, and came to the conclusion that the monitoring facilities of the ProcessMonitor and the ThreadMap are to deeply intertwined with the process control facilities to separate them in a simple manner.

I would now propose a new architecture for the ProcessMonitor which separates the monitoring and the process control contexts.

Here is a roughly overview of what i want to do:

1. Change the internal structure of the ThreadMap; The current ThreadMap is not fit for the task as it also allocates the ThreadMonitor and the ProcessInfos and some of its internal workings are currently dependent on the monitoring part(for example pid retrieval interacts with the ThreadMonitors).I would change the ```map<tid, ThreadMonitor>``` thread map we have currently to a ```map<pid, Process:{ProcessInfo, map<tid, Thread:{ThreadMonitor}>}>``` process map which mirrors the 2-level architecture
we are currently dealing with. The seemingly redundant ```Thread``` class is just there so that the ProcessController mustn't deal with the monitoring-dependent ThreadMonitor. For the new ProcessMap, both the ProcessInfo and the ThreadMonitor would be opaque objects which will be entirely managed by the ProcessMonitor

1. Split up the ProcessMonitor into a ProcessController and a ProcessMonitor, both will be given a reference to the same ProcessMap on construction:
     * ProcessController: This would contain the bulk of the current ProcessMonitor, and would be concerned with the raw managment of the process and all its childs using ptrace. If an event is relevant to the monitoring, the ProcessController calls the relevant callback in the ProcessMonitor. For example if a PTRACE_EVENT_FORK event would occur inside the ProcessController the ProcessController would add the Process and Thread objects for the pid using ProcessMap.insert() and then call monitor_.on_new_process(pid, Process &process) with the newly created Process object and its pid as the argument
     * ProcessMonitor: This class will inherit the monitoring part of the old ProcessMonitor and ThreadMap. It will consist of the different callbacks for the different events in the ProcessController, and will be a child of the MainMonitor class

[...]

PR #54, opened 2018-03-18
Issue 41 -- Handle cmd arguments on lo2s -a 

cvonelm:
This pull requests adds the ability to measure in -a mode as long as a process, either given via the command line and executed by lo2s or as a foreign pid, is running. This goal was achieved by splitting up the ProcessMonitor class into its two components process monitoring and process control ( #41  has a more sincere explanation of what i've done).


src/monitor/process_monitor.cpp
```diff
@@ -20,292 +20,75 @@
...
-        int status;
-        pid_t child = waitpid(-1, &status, __WALL);
-
-        num_wakeups_++;
+    Process &process = process_map().get_process(pid);
 
-        handle_signal(child, status);
+    if(process.info == nullptr)
+    {
+        process.info = new ProcessInfo(pid, false);
```
bmario: `new`

Use std::unique_ptr<...> instead and encapsulate this whole allocation logic inside the respective class.
This is called RAII and is the fundamental principle of resource management in C++

include/lo2s/monitor/dummy_monitor.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <sys/types.h>
+}
+
+namespace lo2s
+{
+
+namespace monitor
+{
+
+class DummyMonitor
```
tilsche: Please split this class in a pure virtual `AbstractProcessMonitor` and a dummy implementation `DummyProcessMonitor`. That way we won't get silent issues if we change/add methods to the base class of the actual `ProcessMonitor`.

tilsche:
Looks good to me, like to wait for bmario's comment next week before the merge.

cvonelm:
Sorry if I understood something wrong there, but currently the calls to make_unique are made inside the ProcessMonitor and not in the Process or Thread class, to ensure a clean separation between the bookkeeping part of the ProcessController and the monitoring part (this is why part of why I have made a Process and Thread class, which are currently somewhat redundant, as they only contain the pid or tid/ppid respectively). I know that this is somewhat hypothetical at the moment but if we would make a new Process Monitoring mode using the existing ThreadMonitor ecosystem (for example a SystemModeThreadMonitor) we would have to make either a new pair of Thread and Process classes which create SystemModeThreadMonitors or have 2 functions inside the Process class (start_process_monitor and start_system_monitor for example), which create the different monitors, which i personally like less than having the ProcessMonitor and SystemModeProcessMonitor create the respective objects for us.

bmario:
Maybe I wasn't clear enough. Let's look for example at the class `Process`. While it is certainly a good idea to have the ProcessInfo member inside of an unique_ptr, I don't see a reason, why this class doesn't provide a member function like this:

```
void initialize_process_info(pid_t pid, bool spawn)
{
    if(!info)
       info = std::make_unique<ProcessInfo>(pid, spawn);
}
```

You're scattering class internals (how is the member initialized), which is hard to maintain.

cvonelm:
Ok, that does make sense to me

cvonelm:
I've came to the conclusion that ProcessMonitor and ProcessController share little to none information, so that the ProcessMap is quite superfluous, and replaced it with a simpler solution where ProcessMonitor and ProcessController do not share their Process and Thread bookkeeping.

PR #55, opened 2018-04-04
Issue 52: Uncaught I/O-exception

phijor:
Previously, if parsing a tracepoint event failed, an uncaught
`std::ios_base::failure` terminated `lo2s`.

Now, a `lo2s::perf::tracepoint::EventFormat::ParseError` is thrown. Alongside,
possible causes for parsing failure are logged, such as missing permissions on
/sys/kernel/debug or a non-existent event.

If parsing fails when opening the tracepoint `SwitchWriter`/`ExitReader`, log
accordingly and die with an (hopefully helpful) error message.  If parsing fails
when opening regular tracepoint events (with `-t`), log and ignore the event.

If the changes are approved of, I'd like to sqash and remove all the typos from
the commit messages.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -50,7 +50,7 @@ int main(int argc, const char** argv)
         if (e.code())
         {
             lo2s::Log::fatal() << "Aborting: " << e.what();
-            return EXIT_FAILURE;
+            return e.code().value();
         }
         else
         {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -49,13 +49,19 @@ int main(int argc, const char** argv)
         // Check for success
         if (e.code())
         {
-            lo2s::Log::error() << "Aborting: " << e.what();
+            lo2s::Log::fatal() << "Aborting: " << e.what();
+            return EXIT_FAILURE;
         }
         else
         {
             lo2s::summary().show();
+            return EXIT_SUCCESS;
         }
-        return e.code().value();
+    }
+    catch (const std::runtime_error& e)
+    {
+        lo2s::Log::fatal() << "Aborting: " << e.what();
+        return EXIT_FAILURE;
     }
 
     return 0;
--- a/src/perf/tracepoint/exit_reader.cpp
+++ b/src/perf/tracepoint/exit_reader.cpp
@@ -45,14 +45,10 @@ ExitReader::ExitReader(int cpu, trace::Trace& trace) try
   comm_field_(get_sched_process_exit_event().field("comm"))
 {
 }
-catch (const EventFormat::Error& e)
+catch (const EventFormat::ParseError& e)
 {
-
-    Log::error() << "Cannot open tracepoint exit reader!";
-    Log::error() << "Failed to open tracepoint event '" << e.event() << "': " << e.what();
-    Log::warn() << "Tracepoint events are inaccessible if read/execute permissions are missing "
-                   "on /sys/kernel/debug";
-    throw std::system_error(EIO, std::system_category());
+    Log::error() << "Failed to open process exit tracepoint event: " << e.what();
+    throw std::runtime_error("Failed to open tracepoint exit reader");
 }
 
 ExitReader::~ExitReader()
--- a/src/perf/tracepoint/exit_reader.cpp
+++ b/src/perf/tracepoint/exit_reader.cpp
@@ -45,6 +45,9 @@ ExitReader::ExitReader(int cpu, trace::Trace& trace) try
   comm_field_(get_sched_process_exit_event().field("comm"))
 {
 }
+// NOTE: function-try-block is intentional; catch get_sched_process_exit_event()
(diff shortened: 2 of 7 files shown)


src/main.cpp
```diff
@@ -49,13 +49,19 @@ int main(int argc, const char** argv)
         // Check for success
         if (e.code())
         {
-            lo2s::Log::error() << "Aborting: " << e.what();
+            lo2s::Log::fatal() << "Aborting: " << e.what();
+            return EXIT_FAILURE;
```
bmario: I think `return e.code().value()` is actually the sensible thing to do here @tilsche ?

src/perf/tracepoint/exit_reader.cpp
```diff
@@ -38,12 +38,18 @@ static const EventFormat& get_sched_process_exit_event()
     return evt;
 }
 
-ExitReader::ExitReader(int cpu, trace::Trace& trace)
-: Reader(cpu, get_sched_process_exit_event().id(), config().mmap_pages), trace_(trace),
+ExitReader::ExitReader(int cpu, trace::Trace& trace) try
```
bmario: If the try-method was intentional, please make a comment, as this syntax is pretty unusual.

src/perf/tracepoint/format.cpp
```diff
@@ -30,32 +30,74 @@
...
 namespace tracepoint
 {
+void EventFormat::parse_id()
+{
+    // Yes, this is C file I/O.  Tell me when you have found a way to easily
+    // access POSIX filesystem errors in C++.
+
+    boost::filesystem::path id_path{ base_path_ / name_ / "id" };
+
+    std::FILE* id_file = std::fopen(id_path.c_str(), "rb");
```
bmario: I have the feeling, this error handling should be a separate function, so we can reuse it.

src/perf/tracepoint/switch_writer.cpp
```diff
@@ -43,14 +43,21 @@ static const EventFormat& get_sched_switch_event()
     return evt;
 }
 
-SwitchWriter::SwitchWriter(int cpu, trace::Trace& trace)
+SwitchWriter::SwitchWriter(int cpu, trace::Trace& trace) try
```
bmario: Please add a comment for the try-method syntax used here

src/perf/tracepoint/format.cpp
```diff
@@ -30,32 +30,74 @@
...
+        throw ParseError{ "Unparseable tracepoint format" };
+    }
 
-    ifs_format.open(path_event / "format");
     std::string line;
-    ifs_format.exceptions(std::ios::badbit);
-    while (getline(ifs_format, line))
+    while (std::getline(ifs_format, line))
     {
+        if (ifs_format.bad())
```
tilsche: That's not going to work. `std::getline` returns the stream itself, so the loop will end the break out of the loop before you can check for `bad()`. Move the `bad()` check to after the loop.

Maybe you could also test for fail() && !eof() to detect potential cases of weird situations.

<rant>Why does typicall `std::getline`-loop always end up with the failbit set? Noone knows... Stupid fstreams are stupid</rant>

src/main.cpp
tilsche: sure

src/perf/tracepoint/format.cpp
tilsche: It may become a bit nifty in collecting / logging the information from the different layers properly, but I agree that it is a good idea.

src/perf/tracepoint/format.cpp
phijor: `std::getline` sets `failbit` on `EOF`. Whether that's good design or not...

src/perf/tracepoint/format.cpp
```diff
@@ -30,34 +30,57 @@
...
+    auto format_path = path_event / "format";
 
-    ifs_id.open(path_event / "id");
+    ifs_id.open(id_path);
     ifs_id >> id_;
 
-    ifs_format.open(path_event / "format");
+    if (ifs_id.fail())
+    {
+        throw ParseError{ "Failed to read tracepoint ID file "s + id_path.string(), errno };
```
bmario: This whole checking is a) not symmetrical for both files and b) looks repetitive. You could refactor it to something like this:

```
read_file_checked(id_path, [this](auto& fs){
    fs >> this->id_;
});

read_file_checked(format_path, [this](auto& fs) {
    std::string line;
    while (std::getline(fs, line)) {
          this->parse_format_line(line);
    }
});
```

bmario: Okay, disregard that, iostreams are broken.
*INSERT YET ANOTHER RANT*

src/main.cpp
bmario: Please change the return code and we're good to go

PR #62, opened 2018-05-28
Sampling in system mode

cvonelm:
This enables sampling in lo2s' system monitoring mode. This is mainly accomplished via a CpuCounterMonitor class which reads sample via a perf::counter::CpuWriter (the perf::counter::Writer was split up into one for sampling in process mode and system mode)


src/summary.cpp
```diff
@@ -58,13 +59,13 @@ std::string pretty_print_bytes(std::size_t trace_size)
 
     std::array<std::string, 5> units = { { "B", "KiB", "MiB", "GiB", "TiB" } };
 
-    while(result_size > 1024)
+    while (result_size > 1024)
```
bmario: The space character after `while` doesn't look like our clang-format style, did you use clang-format?

src/perf/counter/process_writer.cpp
```diff
@@ -0,0 +1,47 @@
...
+{
+namespace counter
+{
+ProcessWriter::ProcessWriter(pid_t pid, pid_t tid, otf2::writer::local& writer,
+                             monitor::MainMonitor& parent, otf2::definition::location scope,
+                             bool enable_on_exec)
+: AbstractWriter(
+      tid, -1, writer,
+      parent.trace().metric_instance(parent.get_metric_class(), writer.location(), scope),
+      enable_on_exec),
```
bmario: This formatting also looks off, did you use clang-format?

src/summary.cpp
cvonelm: Strange, I've used clang-format, and looking at my shells history, the same way I've used it to format all previous changes;

cvonelm:
Please don't merge it currently, I have found a regression

PR #65, opened 2018-06-08
Issue 63 Split up locations to make the creation of writers semantically correct

cvonelm:
This pull request splits up the current locations_ map into 4 different location containers, to ensure uniqueness of the writer and locations where it applies. (For information on which writer/locations are unique and which not, see the comments above the metric_writer, sample_writer and cpu_writer functions inside src/trace/trace.cpp)

### DIFF
--- a/src/trace/trace.cpp
+++ b/src/trace/trace.cpp
@@ -353,41 +353,48 @@ otf2::writer::local& Trace::sample_writer(pid_t pid, pid_t tid)
         std::piecewise_construct, std::forward_as_tuple(tid),
         std::forward_as_tuple(location_ref(), intern(name), location_groups_process_.at(pid),
                               otf2::definition::location::location_type::cpu_thread));
+
     process_comm_groups_.at(pid).add_member(location.first->second);
 
     return archive()(location.first->second);
 }
+
 otf2::writer::local& Trace::cpu_writer(int cpuid)
 {
-    auto name = (boost::format("cpu %d") % cpuid).str();
+    auto name = intern((boost::format("cpu %d") % cpuid).str());
 
     // As CPU IDs are unique in this context, create only one writer/location per
     // CPU ID
     auto r_group = location_groups_cpu_.emplace(
         std::piecewise_construct, std::forward_as_tuple(cpuid),
-        std::forward_as_tuple(location_group_ref(), intern(name),
+        std::forward_as_tuple(location_group_ref(), name,
                               otf2::definition::location_group::location_group_type::unknown,
                               system_tree_cpu_nodes_.at(cpuid)));
+
     const auto& group = r_group.first->second;
 
     auto location = cpu_locations_.emplace(
         std::piecewise_construct, std::forward_as_tuple(cpuid),
-        std::forward_as_tuple(location_ref(), intern(name), group,
+        std::forward_as_tuple(location_ref(), name, group,
                               otf2::definition::location::location_type::cpu_thread));
+
     return archive()(location.first->second);
 }
 
 otf2::writer::local& Trace::metric_writer(pid_t pid, pid_t tid)
 {
     auto name = (boost::format("metrics for thread %d") % tid).str();
+
     // As the tid is unique in this context, create only one
     // writer/location per tid
     auto location = metric_locations_.emplace(
         std::piecewise_construct, std::forward_as_tuple(tid),
         std::forward_as_tuple(location_ref(), intern(name), location_groups_process_.at(pid),
                               otf2::definition::location::location_type::metric));
+
     return archive()(location.first->second);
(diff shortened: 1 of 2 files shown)


src/trace/trace.cpp
```diff
@@ -328,46 +343,56 @@ void Trace::add_lo2s_property(const std::string& name, const std::string& value)
...
 
+    return archive()(location.first->second);
+}
+// As CPU IDs are unique in this context, create only one writer/location per
+// CPU IDs
 otf2::writer::local& Trace::cpu_writer(int cpuid)
 {
-    auto name = intern((boost::format("cpu %d") % cpuid).str());
+    auto name = (boost::format("cpu %d") % cpuid).str();
+    auto intern_name = intern(name);
```
bmario: The new local variable `intern_name` doesn't seem necessary?

src/trace/trace.cpp
```diff
@@ -328,46 +343,56 @@ void Trace::add_lo2s_property(const std::string& name, const std::string& value)
...
-                                       otf2::definition::location::location_type::cpu_thread);
-    return archive()(location);
+    auto location = cpu_locations_.emplace(
+        std::piecewise_construct, std::forward_as_tuple(cpuid),
+        std::forward_as_tuple(location_ref(), intern_name, group,
+                              otf2::definition::location::location_type::cpu_thread));
+    return archive()(location.first->second);
 }
 
+// As pid/tid pairs are unique in this context, create only one
```
bmario: This comment should be inside the function

Also, `tid`alone is already unique.

src/trace/trace.cpp
```diff
@@ -328,46 +343,56 @@ void Trace::add_lo2s_property(const std::string& name, const std::string& value)
...
-                           otf2::definition::location::location_type::metric);
-    return archive()(location);
+    auto location = metric_locations_.emplace(
+        std::piecewise_construct, std::forward_as_tuple(tid),
+        std::forward_as_tuple(location_ref(), intern(name), location_groups_process_.at(pid),
+                              otf2::definition::location::location_type::metric));
+    return archive()(location.first->second);
 }
-
+// As names may not be unique in this context, always generate a new location/writer pair
```
bmario: This comment should be inside the function

src/trace/trace.cpp
```diff
@@ -332,42 +347,53 @@ void Trace::add_lo2s_property(const std::string& name, const std::string& value)
...
+        std::forward_as_tuple(location_ref(), intern(name), location_groups_process_.at(pid),
+                              otf2::definition::location::location_type::cpu_thread));
+    process_comm_groups_.at(pid).add_member(location.first->second);
+
+    return archive()(location.first->second);
+}
 otf2::writer::local& Trace::cpu_writer(int cpuid)
 {
-    auto name = intern((boost::format("cpu %d") % cpuid).str());
+    auto name = (boost::format("cpu %d") % cpuid).str();
```
bmario: Maybe I was a bit unclear, but you got me wrong on this one. I actually meant to use the old code snippet, which does the `intern()` call once. While the current code should work fine 
as `intern()` does caching, it's less efficient than reusing the interned string:

```
auto name = intern((boost::format("cpu %d") % cpuid).str());
```

src/trace/trace.cpp
```diff
@@ -332,42 +347,53 @@ void Trace::add_lo2s_property(const std::string& name, const std::string& value)
...
+    // As the tid is unique in this context, create only one
+    // writer/location per tid
+    auto location = metric_locations_.emplace(
+        std::piecewise_construct, std::forward_as_tuple(tid),
+        std::forward_as_tuple(location_ref(), intern(name), location_groups_process_.at(pid),
+                              otf2::definition::location::location_type::metric));
+    return archive()(location.first->second);
 }
-
 otf2::writer::local& Trace::metric_writer(const std::string& name)
```
bmario: Please give the code some space to breath :) (aka insert a newline between the two functions)

src/trace/trace.cpp
```diff
@@ -332,42 +347,53 @@ void Trace::add_lo2s_property(const std::string& name, const std::string& value)
...
 
+    // As the tid is unique in this context, create only one writer/location per tid
+    auto location = thread_locations_.emplace(
+        std::piecewise_construct, std::forward_as_tuple(tid),
+        std::forward_as_tuple(location_ref(), intern(name), location_groups_process_.at(pid),
+                              otf2::definition::location::location_type::cpu_thread));
+    process_comm_groups_.at(pid).add_member(location.first->second);
+
+    return archive()(location.first->second);
+}
```
bmario: Please give the code some space to breath :) (aka insert a newline between the two functions)

Issue #64, opened 2018-06-08
Add support for raw metrics


rschoene:
In `perf`, I can pass raw metrics via the `-e` flag. This should also be supported by lo2s for `-e` and `-E`

Raw metrics are encoded in the following way `rNNNN`. More information is given in `man perf-list`

cvonelm:
I have made a first implementation of this on the issue-64 branch. and tried it with lo2s -E r412e firestarter (r412e being the alias for the 'ram' event) and it produced results which where pretty much indistinguishable from using the 'ram' event directly, however before we go on, I would like to test this out on a more unorthodox setting(one where using a hexadecimal identifier is actually useful, because there is no human-readable identifier), is there a specific event and its hexadecimal identifier, which would be useful/interesting to test out on one of our test systems?

bmario:
Fixed in 2c1514b2790b3f683f9dd4e4139d692748bbf137

PR #66, opened 2018-06-12
raw events

tilsche:
for #64

### DIFF
--- a/src/perf/event_provider.cpp
+++ b/src/perf/event_provider.cpp
@@ -138,7 +138,7 @@ constexpr std::uint64_t bit(int bitnumber)
 {
     return static_cast<std::uint64_t>(1_u64 << bitnumber);
 }
-} // namespace
+}
 
 namespace lo2s
 {
@@ -489,7 +489,7 @@ EventProvider::EventProvider()
 const CounterDescription& EventProvider::cache_event(const std::string& name)
 {
     // Format for raw events is rNNNN
-    static const std::regex raw_regex("r[[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]]");
+    static const std::regex raw_regex("r[[:xdigit:]]{4}");
 
     // save event in event map; return a reference to the inserted event to
     // the caller.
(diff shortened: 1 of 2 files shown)


src/perf/event_provider.cpp
```diff
@@ -475,12 +488,23 @@ EventProvider::EventProvider()
 
 const CounterDescription& EventProvider::cache_event(const std::string& name)
 {
+    // Format for raw events is rNNNN
+    static const std::regex raw_regex("r[[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]]");
```
tilsche: Please use `[[:xdigit:]]{1-16}`.

src/perf/event_provider.cpp
```diff
@@ -137,7 +138,7 @@ constexpr std::uint64_t bit(int bitnumber)
 {
     return static_cast<std::uint64_t>(1_u64 << bitnumber);
 }
-}
+} // namespace
```
tilsche: :-)

src/perf/event_provider.cpp
```diff
@@ -345,6 +346,18 @@ static void event_description_update(CounterDescription& event, std::uint64_t va
...
 }
 
+const CounterDescription raw_read_event(const std::string& ev_desc)
+{
+    uint64_t code = std::stoull(ev_desc.substr(1), nullptr, 16);
+    const CounterDescription event(ev_desc, PERF_TYPE_RAW, code, 0);
+
+    if (!event_is_openable(event))
+    {
+        throw EventProvider::InvalidEvent("cannot open requested event");
```
bmario: Add the name of the event, which failed

cvonelm: Is that needed? The error message in that case ( Example: " 'r003c' does not name a known event, ignoring! (reason: cannot open requested event)") already contains the name of the event

bmario: Oh, it get's added somewhere higher in call stack?

cvonelm: Seems like that

PR #70, opened 2018-07-02
Remove default metrics, dont record if no valid metrics are given

cvonelm:
This removes the default set of metrics, as well as not starting to record metrics if no valid (non-leader) events are given. If none are given in process mode lo2s will exit. If none are given in cpu mode lo2s will just not record any metrics.

### DIFF
--- a/src/monitor/main_monitor.cpp
+++ b/src/monitor/main_monitor.cpp
@@ -88,12 +88,6 @@ otf2::definition::metric_class MainMonitor::generate_metric_class()
                                           otf2::common::metric_mode::accumulated_start,
                                           otf2::common::type::Double, "#"));
     }
-    c.add_member(trace_.metric_member("time_enabled", "time event active",
-                                      otf2::common::metric_mode::accumulated_start,
-                                      otf2::common::type::uint64, "ns"));
-    c.add_member(trace_.metric_member("time_running", "time event on CPU",
-                                      otf2::common::metric_mode::accumulated_start,
-                                      otf2::common::type::uint64, "ns"));
     return c;
 }
 MainMonitor::~MainMonitor()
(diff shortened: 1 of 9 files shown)


src/monitor/main_monitor.cpp
```diff
@@ -88,12 +88,6 @@ otf2::definition::metric_class MainMonitor::generate_metric_class()
                                           otf2::common::metric_mode::accumulated_start,
                                           otf2::common::type::Double, "#"));
     }
-    c.add_member(trace_.metric_member("time_enabled", "time event active",
-                                      otf2::common::metric_mode::accumulated_start,
-                                      otf2::common::type::uint64, "ns"));
-    c.add_member(trace_.metric_member("time_running", "time event on CPU",
-                                      otf2::common::metric_mode::accumulated_start,
-                                      otf2::common::type::uint64, "ns"));
```
phijor: Are you sure we want to remove `time_enabled` and `time_running` by default? These aren't really metrics per se, but their values are updated from the kernel buffer every time the metric events are read. I think it would make sense to stop recording them if there's no metric events being recorded, not just outright disable them.

bmario: What is the point of those metrics, when you're not recording any metrics? What would their values even be?

The correct behavior:
- [ ] if there are any perf metrics recorded, add those pseudo metrics
- [ ] if there are not metrics recorded, either the requested none or requested invalid metrics, then don't record them

The same goes for the metric leader.

src/monitor/process_monitor.cpp
```diff
@@ -23,13 +23,19 @@
...
 namespace lo2s
 {
 namespace monitor
 {
 
 ProcessMonitor::ProcessMonitor() : MainMonitor()
 {
+    if (perf::requested_events().events.size() == 0)
+    {
+        throw std::runtime_error("No valid metric events given");
```
bmario: Why exactly is this an exception? I want to be able to use lo2s without the need for recording any metrics

PR #76, opened 2018-07-10
Determine default metric leader event at runtime

phijor:


CMakeLists.txt
```diff
@@ -269,14 +269,6 @@ else()
     endif()
 endif()
 
-message(STATUS "Linux kernel version: ${LINUX_VERSION}")
```
bmario: This is still a fair message to print in cmake

src/config.cpp
```diff
@@ -234,6 +234,8 @@ const Config& config()
 }
 void parse_program_options(int argc, const char** argv)
 {
+    lo2s::logging::set_min_severity_level(nitro::log::severity_level::error);
```
bmario: This looks like either unrelated or an ugly hack :(

src/perf/event_provider.cpp
```diff
@@ -254,6 +254,29 @@ std::vector<std::string> EventProvider::get_pmu_event_names()
     return names;
 }
 
+const CounterDescription& EventProvider::get_default_metric_leader_event()
+{
+    Log::debug() << "checking for metric leader event...";
+    for (auto candidate : {
+             "ref-cycles",
+             "bus-cycles",
```
tilsche: Add `cpu-cycles` for EPYC.

src/perf/event_provider.cpp
```diff
@@ -254,6 +254,29 @@ std::vector<std::string> EventProvider::get_pmu_event_names()
     return names;
 }
 
+const CounterDescription& EventProvider::get_default_metric_leader_event()
+{
+    Log::debug() << "checking for metric leader event...";
+    for (auto candidate : {
+             "ref-cycles",
```
bmario: cpu-cycles is missing

src/config.cpp
```diff
@@ -364,7 +366,7 @@ void parse_program_options(int argc, const char** argv)
         ("metric-leader",
             po::value(&config.metric_leader)
                 ->value_name("EVENT")
-                ->default_value(QUOTE(DEFAULT_METRIC_LEADER)),
+                ->default_value(perf::EventProvider::get_default_metric_leader_event().name),
```
bmario: On a second thought, I'm not convinced that this is a good idea. Using perf in program options arg parsing has some problems, e.g.: perf might not be available, there might not be any suitable metric leaders, it might cause some latency, which may be a bad idea for shell autocompletion, ...

Instead, remove the default_value and either use the given metric_leader, or determine a possible metric leader later on.

PR #77, opened 2018-07-11
Use improved otf2xx metric event handling

bmario:
This should improve performance for writing any metric to otf2 traces

### DIFF
--- a/src/metric/plugin/channel.cpp
+++ b/src/metric/plugin/channel.cpp
@@ -34,29 +34,31 @@ namespace metric
 namespace plugin
 {
 
+static otf2::definition::metric_instance&&
+make_metric_instance(otf2::definition::metric_instance&& instance,
+                     otf2::definition::metric_member&& member)
+{
+    auto mc = otf2::definition::make_weak_ref(instance.metric_class());
+    mc->add_member(std::move(member));
+    return std::move(instance);
+}
+
+static const char* empty_if_null(const char* cstr)
+{
+    return cstr == nullptr ? "" : cstr;
+}
+
 Channel::Channel(const char* name, const char* description, const char* unit, wrapper::Mode mode,
                  wrapper::ValueType value_type, trace::Trace& trace)
-: id_(-1), name_(name), description_(), unit_(), mode_(mode), value_type_(value_type),
-  writer_(trace.metric_writer(name_)),
-  metric_(trace.metric_instance(trace.metric_class(), writer_.location(),
-                                writer_.location().location_group().parent()))
+: id_(-1), name_(name), description_(empty_if_null(description)), unit_(empty_if_null(unit)),
+  mode_(mode), value_type_(value_type), writer_(trace.metric_writer(name_)),
+  metric_(
+      make_metric_instance(trace.metric_instance(trace.metric_class(), writer_.location(),
+                                                 writer_.location().location_group().parent()),
+                           trace.metric_member(name_, description_, wrapper::convert_mode(mode_),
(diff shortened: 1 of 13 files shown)


src/metric/plugin/channel.cpp
```diff
@@ -34,29 +34,31 @@ namespace metric
...
+Channel::Channel(const char* name, const char* description, const char* unit, wrapper::Mode mode,
+                 wrapper::ValueType value_type, trace::Trace& trace)
+: id_(-1), name_(name), description_(empty_if_null(description)), unit_(empty_if_null(unit)),
+  mode_(mode), value_type_(value_type), writer_(trace.metric_writer(name_)),
+  metric_(
+      make_metric_instance(trace.metric_instance(trace.metric_class(), writer_.location(),
+                                                 writer_.location().location_group().parent()),
+                           trace.metric_member(name_, description_, wrapper::convert_mode(mode_),
+                                               wrapper::convert_type(value_type_), unit_))),
+  event_(otf2::chrono::genesis(), metric_.metric_class())
```
bmario: the event needs to reference the metric instance, not the class

src/metric/x86_energy/monitor.cpp
```diff
@@ -19,10 +19,10 @@ Monitor::Monitor(::x86_energy::SourceCounter counter, int cpu,
                  const otf2::definition::system_tree_node& stn)
 : IntervalMonitor(trace, std::to_string(cpu), sampling_interval), counter_(std::move(counter)),
   cpu_(cpu), otf2_writer_(trace.metric_writer(name())),
-  metric_instance_(trace.metric_instance(metric_class, otf2_writer_.location(), stn))
+  metric_instance_(trace.metric_instance(metric_class, otf2_writer_.location(), stn)),
+  metric_event_(otf2::chrono::genesis(), metric_instance)
```
bmario: This can't possibly compile, should be `metric_instance_`

src/perf/tracepoint/writer.cpp
```diff
@@ -20,31 +20,31 @@ Writer::Writer(int cpu, const EventFormat& event, trace::Trace& trace_,
   writer_(trace_.metric_writer((boost::format("tracepoint metrics for CPU %d") % cpu).str())),
   metric_instance_(
       trace_.metric_instance(metric_class, writer_.location(), trace_.system_tree_cpu_node(cpu))),
-  time_converter_(perf::time::Converter::instance())
+  time_converter_(perf::time::Converter::instance()),
+  metric_event_(otf2::chrono::genesis(), metric_class)
```
bmario: the event needs to reference the metric instance, not the class

src/perf/tracepoint/writer.cpp
```diff
@@ -20,31 +20,31 @@ Writer::Writer(int cpu, const EventFormat& event, trace::Trace& trace_,
...
 }
 
 bool Writer::handle(const Reader::RecordSampleType* sample)
 {
     auto tp = time_converter_(sample->time);
 
-    size_t index = 0;
-    for (const auto& field : event_.fields())
+    metric_event_.timestamp(tp);
+    for (const auto& index_field : nitro::lang::enumerate(event_.fields()))
```
bmario: You can't use `nitro::lang::enumerate` here.
In case of: `field.is_integer()` the indices will be shifted. Not good.

bmario: There even has been a commit in the past changing it from enumerate to this ordinary loop:
70d14732f4ea53c28e06d1856ba2e5ba4833fba3

PR #82, opened 2018-08-22
Issue 9 sampling system mode

cvonelm:
This pull request enables sampling in system mode, currently hidden behind the --enable-system-sampling Flag. Sampling and switching events are currently recorded in different writers, as unifying the timepoints between the SwitchWriter and the SampleWriter seems to be fairly non trivial(Could become trivial with  PERF_RECORD_SWITCH). Currently if samples are recorded in system mode, nothing is displayed for the switch events in Vampir's timeline. This seems to be most likely a bug in Vampir, as the missing event records show up when manually investigating the trace with otf2-print


include/lo2s/monitor/main_monitor.hpp
```diff
@@ -55,9 +69,17 @@ class MainMonitor
         return trace_;
     }
 
+    void insert_cached_mmap_events(std::deque<struct MmapCache> cached_events);
```
bmario: No need for struct.
Why no `const& `?

src/perf/sample/writer.cpp
```diff
@@ -90,57 +90,53 @@ Writer::~Writer()
...
     {
-        const auto& mapping =
-            trace_.merge_ips(local_ip_refs_, ip_ref_counter_, monitor_.info().maps());
+        const auto& mapping = trace_.merge_ips(local_ip_refs_, monitor_.get_process_infos());
         otf2_writer_ << mapping;
     }
 }
-
-trace::IpRefMap::iterator Writer::find_ip_child(Address addr, trace::IpRefMap& children)
+trace::IpRefMap::iterator Writer::find_ip_child(Address addr, pid_t pid, trace::IpRefMap& children)
```
bmario: Please readd the new line above the function

include/lo2s/trace/trace.hpp
```diff
@@ -93,11 +95,12 @@ class Trace
...
     void process_update_executable(pid_t pid, const std::string& exe_name);
 
-    otf2::writer::local& sample_writer(pid_t pid, pid_t tid);
-    otf2::writer::local& cpu_writer(int cpuid);
-    otf2::writer::local& metric_writer(pid_t pid, pid_t tid);
-    otf2::writer::local& metric_writer(const std::string& name);
+    otf2::writer::local& thread_sample_writer(pid_t pid, pid_t tid);
+    otf2::writer::local& cpu_sample_writer(int cpuid);
+    otf2::writer::local& thread_metric_writer(pid_t pid, pid_t tid);
+    otf2::writer::local& named_metric_writer(const std::string& name);
```
bmario: Please verify that you caught all instances, where you need to change that name. I think there is one missing for metric/x86_energy.

bmario:
Please provide a small example trace file, so we can open a bug in the Vampir issue tracker.

cvonelm:
[minimal_trace.zip](https://github.com/tud-zih-energy/lo2s/files/2328391/minimal_trace.zip)
New trace with unknown replaced with process, the vampir warnings did go away. but there is still nothing showing up for switch events

cvonelm:
[minimal_trace.zip](https://github.com/tud-zih-energy/lo2s/files/2345640/minimal_trace.zip)
Every ENTER is now always closed with a LEAVE, but the issue remains

bmario:
To quote @JoZie:

>The problem is, that your Trace contains samples and "Plain" enter/leave events, so Vampir expects a sampling trace.
This means, that it is expecting "CallingContext" enter/leave events instead of the "Plain" ones.
I'm not certain if your combination of "Plain" and "CallingContext" events is allowed with OTF2.

That is a big bummer because otf2xx doesn't know about CallingContextEnter and CallingContextLeave. We may have to proceed with #84 first, so we have an up-to-date otf2xx in master. Then, implement these in otf2xx, update otf2xx here and finally merge this.

src/config.cpp
```diff
@@ -311,6 +311,9 @@ void parse_program_options(int argc, const char** argv)
                 ->value_name("N")
                 ->default_value(11010113),
             "Sampling period (in number of events specified by -e).")
+        ("enable-system-sampling",
```
bmario: There is no other option with the prefix enable, so I'd prefer `--system-sampling` with the short option `-A`

src/config.cpp
```diff
@@ -311,6 +311,9 @@ void parse_program_options(int argc, const char** argv)
...
+            po::bool_switch(&config.system_mode_sampling),
+            "Enable sampling in system mode")
```
bmario: The description reads `Enable sampling in system mode`, but a check (which I'd like more) or implicit enabling of system monitoring isn't done.
Aka. this does not fail with an error:
```
$ lo2s --enable-system-sampling
```

bmario:
I discussed it with @tilsche today, and we want a slightly different setup for the command line arguments.

- There is the already existing `-a` to toggle between system-wide mode and process mode
- There is a new option `--instruction-sampling` (and in a later PR `--no-instruction-sampling`). These arguments enabled and disable the instruction sampling independently of the mode
- There is a second new option `-A`, which is a short-hand for `-a --instruction-sampling`

This way, we can request all four combinations of sampling enabled/disabled and system/process mode:

Arguments  | Sampling/Mode
------------ | -------------
No arguments or `--instruction-sampling` | *process mode with instruction-sampling*
`--no-instruction-sampling` | *process mode without instruction-sampling*
`-a` | *system mode without instruction-sampling*
`-a --instruction-sampling` or short `-A` | *system mode with instruction-sampling*

PR #84, opened 2018-08-27
Issue 78: use PERF_RECORD_COMM events to trace task command changes

phijor:

PR #89, opened 2018-09-26
Adds perf record switch support

### LINKED ISSUE #79: Look into PERF_RECORD_SWITCH
There are relatively new (Linux 4.3) records which seem to provide the same information as the  `sched_switch` tracepoint.

> This functionality is in addition to tracepoint and software events for measuring context switches.  The advantage of this method is that it **will full information even with strict `perf_event_paranoid` settings**.

We should also look into that if it gives other advantages, e.g. less overhead than tracepoints.

cvonelm:
This fixes #79.

This PR implements recording of ContextSwitch Events using PERF_RECORD_SWITCH instead of the older tracepoint based SwitchWriter, which has the advantage of working without root privileges. 

The only remaining thing that needs root privileges in system-wide mode is the ExitReader, which we should replace shortly with a PERF_RECORD_COMM based solution similar to #84

As PERF_RECORD_SWITCH events were only introduced in Kernel 4.3, use of this feature is guarded by a cmake check and the cmake option USE_PERF_RECORD_SWITCH

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,7 +107,7 @@ CHECK_STRUCT_HAS_BITFIELD("struct perf_event_attr" context_switch linux/perf_eve
 CMAKE_DEPENDENT_OPTION(USE_RADARE "Enable Radare support." ON "Radare_FOUND" OFF)
 option(HW_BREAKPOINT_COMPAT "Time synchronization fallback for old kernels without hardware breakpoint support." OFF)
 option(USE_PERF_CLOCKID "Enables specifying a custom reference clock for recorded events" ON)
-option(USE_PERF_RECORD_SWITCH "Uses PERF_RECORD_SWITCH for CPU Context Switches instead of the older tracepoint based solution" ${HAVE_PERF_RECORD_SWITCH})
+CMAKE_DEPENDENT_OPTION(USE_PERF_RECORD_SWITCH "Uses PERF_RECORD_SWITCH for CPU Context Switches instead of the older tracepoint based solution" ON HAVE_PERF_RECORD_SWITCH OFF)
 option(IWYU "Developer option for include what you use." OFF)
 option(UML_LOOK "Generate graphs with an UML look" OFF)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,7 @@ option(UML_LOOK "Generate graphs with an UML look" OFF)
 # system configuration checks
 CHECK_INCLUDE_FILES(linux/hw_breakpoint.h HAVE_HW_BREAKPOINT_H)
 CHECK_STRUCT_HAS_MEMBER("struct perf_event_attr" clockid linux/perf_event.h HAVE_PERF_EVENT_ATTR_CLOCKID)
+CHECK_STRUCT_HAS_MEMBER("struct perf_event_attr" context_switch linux/perf_event.h HAVE_PERF_RECORD_SWITCH)
 check_function_exists(clock_gettime CLOCK_GETTIME_FOUND)
 if(NOT CLOCK_GETTIME_FOUND)
     set(CMAKE_REQUIRED_LIBRARIES "rt")
@@ -145,10 +146,9 @@ set(SOURCE_FILES
     src/perf/counter/abstract_writer.cpp
 
     src/perf/sample/writer.cpp
-    src/perf/context_switch/writer.cpp
     src/perf/time/converter.cpp src/perf/time/reader.cpp
     src/perf/tracepoint/format.cpp src/perf/tracepoint/metric_monitor.cpp
-    src/perf/tracepoint/writer.cpp src/perf/tracepoint/switch_writer.cpp
+    src/perf/tracepoint/writer.cpp
     src/perf/tracepoint/exit_reader.cpp
 
     src/time/time.cpp
@@ -270,6 +270,13 @@ else()
     if (HAVE_PERF_EVENT_ATTR_CLOCKID)
         message(WARNING "The system claims to support custom reference clocks for perf events. Consider activating USE_PERF_CLOCKID to enable the options -k,--clockid.")
     endif()
+endif()
+
+if(HAVE_PERF_RECORD_SWITCH)
+    target_compile_definitions(lo2s PRIVATE USE_PERF_RECORD_SWITCH)
+    target_sources(lo2s PRIVATE src/perf/context_switch/writer.cpp)
+else()
+    target_sources(lo2s PRIVATE src/perf/tracepoint/switch_writer.cpp)
 endif()
 
 message(STATUS "Linux kernel version: ${LINUX_VERSION}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -275,7 +275,7 @@ else()
     endif()
 endif()
 
-if( USE_PERF_RECORD_SWITCH)
+if(USE_PERF_RECORD_SWITCH)
     target_compile_definitions(lo2s PRIVATE USE_PERF_RECORD_SWITCH)
     target_sources(lo2s PRIVATE src/perf/context_switch/writer.cpp)
 else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,6 +145,7 @@ set(SOURCE_FILES
     src/perf/counter/abstract_writer.cpp
 
     src/perf/sample/writer.cpp
+    src/perf/context_switch/writer.cpp
     src/perf/time/converter.cpp src/perf/time/reader.cpp
     src/perf/tracepoint/format.cpp src/perf/tracepoint/metric_monitor.cpp
     src/perf/tracepoint/writer.cpp src/perf/tracepoint/switch_writer.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,7 @@ SET(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake;${CMAKE_MODULE_PATH}")
 
 include(cmake/DefaultBuildType.cmake)
 include(cmake/UnsetIfUpdated.cmake)
+include(cmake/CheckStructHasBitField.cmake)
 
 # Intialize git submodules if not done already
 include(cmake/GitSubmoduleUpdate.cmake)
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -100,10 +101,13 @@ find_package(Threads REQUIRED)
 find_package(Doxygen COMPONENTS dot)
 find_package(x86_energy 2.0 CONFIG)
 
+CHECK_STRUCT_HAS_BITFIELD("struct perf_event_attr" context_switch linux/perf_event.h HAVE_PERF_RECORD_SWITCH)
+
 # configurable options
 CMAKE_DEPENDENT_OPTION(USE_RADARE "Enable Radare support." ON "Radare_FOUND" OFF)
 option(HW_BREAKPOINT_COMPAT "Time synchronization fallback for old kernels without hardware breakpoint support." OFF)
 option(USE_PERF_CLOCKID "Enables specifying a custom reference clock for recorded events" ON)
+option(USE_PERF_RECORD_SWITCH "Uses PERF_RECORD_SWITCH for CPU Context Switches instead of the older tracepoint based solution" ${HAVE_PERF_RECORD_SWITCH})
```
bmario: This should be a `CMAKE_DEPENDENT_OPTION`

cmake/CheckStructHasBitField.cmake
```diff
@@ -0,0 +1,20 @@
...
+    int main()
+    {
+        ${_STRUCT} test;
+        test.${_MEMBER} = 0;
+        return 0;
+    }
+    ")
+
+    CHECK_C_SOURCE_COMPILES("${_CHECK_STRUCT_MEMBER_SOURCE_CODE}" ${_RESULT})
+endmacro()
```
bmario: Unset local variables create with set

src/perf/context_switch/writer.cpp
```diff
@@ -0,0 +1,100 @@
...
+Writer::Writer(int cpu, trace::Trace& trace)
+: Reader(cpu), otf2_writer_(trace.cpu_switch_writer(cpu)), time_converter_(time::Converter::instance()),
+  trace_(trace)
+{
+}
+
+Writer::~Writer()
+{
+    // Always close the last entered region
+    if (last_event_type == ENTER)
```
bmario:
```suggestion
    if (last_event_type_ == LastEventType::enter)
```

src/perf/context_switch/writer.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    auto tp = time_converter_(context_switch->time);
+    auto elem = thread_calling_context_refs_.emplace(std::piecewise_construct,
+                                            std::forward_as_tuple(context_switch->next_prev_pid),
+                                            std::forward_as_tuple(thread_calling_context_refs_.size()));
+
+    // Check if we left the region
+    if (context_switch->header.misc & PERF_RECORD_MISC_SWITCH_OUT)
+    {
+        if(last_event_type == LEAVE)
```
bmario:
```suggestion
        // we leave the region
        if(last_event_type_ == LastEventType::leave)
```

src/perf/context_switch/writer.cpp
```diff
@@ -0,0 +1,100 @@
...
+        {
+            otf2_writer_.write_calling_context_enter(tp, elem.first->second, 2);
+        }
+        otf2_writer_.write_calling_context_leave(tp, elem.first->second);
+        last_event_type = LEAVE;
```
bmario:
```suggestion
        last_event_type_ = LastEventType::leave;
```

src/perf/context_switch/writer.cpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+    else
+    {
+        if(last_event_type == ENTER)
```
bmario:
```suggestion
        if(last_event_type_ == LastEventType::enter)
```

src/perf/context_switch/writer.cpp
```diff
@@ -0,0 +1,100 @@
...
+        {
+            otf2_writer_.write_calling_context_leave(tp, last_pid_);
+        }
+        otf2_writer_.write_calling_context_enter(tp, elem.first->second, 2);
+
+        summary().register_process(context_switch->next_prev_pid);
+        last_event_type = ENTER;
```
bmario:
```suggestion
        last_event_type_ = LastEventType::enter;
```

include/lo2s/perf/context_switch/writer.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+    pid_t last_pid_;
+    otf2::chrono::time_point last_time_point_ = otf2::chrono::genesis();
+
+    enum class LastEventType
+    {
+        enter,
+        leave
+    };
+    LastEventType last_event_type = LastEventType::leave;
```
bmario:
```suggestion
    LastEventType last_event_type_ = LastEventType::leave;
```
Missing `_` postfix for member variable

src/perf/context_switch/writer.cpp
```diff
@@ -0,0 +1,100 @@
...
+            otf2_writer_.write_calling_context_enter(tp, elem.first->second, 2);
+        }
+        otf2_writer_.write_calling_context_leave(tp, elem.first->second);
+        last_event_type_ = LastEventType::leave;
+    }
+    else
+    {
+        if(last_event_type_ == LastEventType::enter)
+        {
+            otf2_writer_.write_calling_context_leave(tp, last_pid_);
```
bmario: You're using a `pid_t` here, not a `otf2::definition::calling_context::reference_type`.
`last_pid_` has to be looked up in the `thread_calling_context_refs_` map.

src/perf/context_switch/writer.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+}
+
+Writer::~Writer()
+{
+    // Always close the last entered region
+    if (last_event_type_ == LastEventType::enter)
+    {
+        auto elem =
+            thread_calling_context_refs_.emplace(std::piecewise_construct, std::forward_as_tuple(last_pid_),
```
bmario: This emplace should never actually perform an emplace. `last_pid_` should already exists.

include/lo2s/perf/context_switch/writer.hpp
```diff
@@ -0,0 +1,64 @@
...
+    using Reader<Writer>::handle;
+    bool handle(const Reader::RecordSwitchCpuWideType* context_switch);
+
+private:
+    otf2::writer::local& otf2_writer_;
+    const time::Converter time_converter_;
+    trace::Trace& trace_;
+    std::unordered_map<pid_t, otf2::definition::calling_context::reference_type> thread_calling_context_refs_;
+
+    pid_t last_pid_;
```
bmario: `last_pid_` can actually be a `otf2::definition::calling_context::reference_type`

src/perf/context_switch/writer.cpp
```diff
@@ -0,0 +1,96 @@
...
+    auto elem = thread_calling_context_refs_.emplace(std::piecewise_construct,
+                                            std::forward_as_tuple(context_switch->next_prev_pid),
+                                            std::forward_as_tuple(thread_calling_context_refs_.size()));
+
+    // Check if we left the region
+    if (context_switch->header.misc & PERF_RECORD_MISC_SWITCH_OUT)
+    {
+        if(last_event_type_ == LastEventType::leave)
+        {
+            otf2_writer_.write_calling_context_enter(tp, elem.first->second, 2);
```
bmario: This branch means that perf is saying, we have left a region, but for some reason, we didn't catch the enter.
How common is that case? Please add a log::trace() for this situation

@tilsche Shouldn't the assumption, in this case, be that the region was entered at the last known time point?

src/perf/context_switch/writer.cpp
```diff
@@ -0,0 +1,96 @@
...
+        }
+        otf2_writer_.write_calling_context_leave(tp, elem.first->second);
+        last_event_type_ = LastEventType::leave;
+    }
+    else
+    {
+        if(last_event_type_ == LastEventType::enter)
+        {
+            otf2_writer_.write_calling_context_leave(tp, elem.first->second);
```
bmario: This means we enter a region but somehow missed the leave of the current.
How common is that case? Please add a log::trace() for this situation

src/perf/context_switch/writer.cpp
cvonelm: Afaik the only instance where we are loosing enter or leaves is when we are loosing samples due to high system load. But this way we always maintain consistency.

Also vampir segfaults when given a trace where calling context events are not consistently closed/opened

cvonelm: [trace_segfault.zip](https://cloudstore.zih.tu-dresden.de/index.php/s/taS7Z1lNqr3a637) Here is a trace where I simulated this situation by randomly dropping some of the enter/leaves if you want to report this to the Vampir team

bmario: Yeah sure, I'm just wondering about the timestamp the enter should have.
Because the obvious case is, when we start recording a process is already scheduled at that time. Then, we'll only see the leave

src/perf/context_switch/writer.cpp
```diff
@@ -0,0 +1,95 @@
...
+namespace lo2s
+{
+namespace perf
+{
+namespace context_switch
+{
+
+Writer::Writer(int cpu, trace::Trace& trace)
+: Reader(cpu), otf2_writer_(trace.cpu_switch_writer(cpu)), time_converter_(time::Converter::instance()),
+  trace_(trace)
```
bmario:
```suggestion
  trace_(trace), last_time_point_(lo2s::time::now())
```

src/perf/context_switch/writer.cpp
```diff
@@ -0,0 +1,95 @@
...
+    auto tp = time_converter_(context_switch->time);
+    auto elem = thread_calling_context_refs_.emplace(std::piecewise_construct,
+                                            std::forward_as_tuple(context_switch->next_prev_pid),
+                                            std::forward_as_tuple(thread_calling_context_refs_.size()));
+    if (context_switch->header.misc & PERF_RECORD_MISC_SWITCH_OUT)
+    {
+        if(last_event_type_ == LastEventType::leave)
+        {
+            Log::trace() << "Writing missing calling context enter event for current calling context leave event.";
+            otf2_writer_.write_calling_context_enter(last_time_point_, elem.first->second, 2);
```
bmario:
```suggestion
            otf2_writer_.write_calling_context_enter(std::min(last_time_point_, tp), elem.first->second, 2);
```

cvonelm:
The problem is that switch is a reserved keyword

bmario:
Doh. You're right. This is why we can't have good things.

PR #94, opened 2018-10-07
Write program begin/end records for processes

### LINKED ISSUE #80: Look into ProgramBegin/ProgramEnd OTF2 records
https://silc.zih.tu-dresden.de/otf2-current/html/group__records__event.html#ProgramBegin

Sounds like a good addition to process mode.

phijor:
Fixes #80

### DIFF
--- a/include/lo2s/trace/trace.hpp
+++ b/include/lo2s/trace/trace.hpp
@@ -34,6 +34,7 @@
 #include <map>
 #include <mutex>
 #include <unordered_map>
+#include <unordered_set>
 
 namespace lo2s
 {
@@ -98,13 +99,16 @@ private:
     void add_thread_exclusive(pid_t tid, const std::string& name,
                               const std::lock_guard<std::mutex>&);
 
+    void mark_process_begin(pid_t pid, const otf2::definition::string& prog_name,
+                            const std::vector<std::string>& command_line);
+
     void update_process_name(pid_t pid, const otf2::definition::string& name);
     void update_thread_name(pid_t tid, const otf2::definition::string& name);
 
 public:
     void add_cpu(int cpuid);
 
-    void add_process(pid_t pid, pid_t parent, const std::string& name = "");
+    void add_process(pid_t pid, pid_t parent, const std::vector<std::string>& command_line);
 
     void add_thread(pid_t tid, const std::string& name);
     void add_threads(const std::unordered_map<pid_t, std::string>& tid_map);
@@ -113,6 +117,8 @@ public:
 
     void update_process_name(pid_t pid, const std::string& name);
     void update_thread_name(pid_t tid, const std::string& name);
+
+    void mark_process_end(pid_t pid, int status_code);
 
     otf2::writer::local& thread_sample_writer(pid_t pid, pid_t tid);
     otf2::writer::local& cpu_sample_writer(int cpuid);
@@ -299,6 +305,14 @@ private:
     std::map<int, otf2::definition::location> cpu_metric_locations_;
     std::map<int, otf2::definition::location> cpu_switch_locations_;
     otf2::definition::container<otf2::definition::location> named_metric_locations_;
+
+    // For each program_begin event written to the thread_location belonging to
+    // <pid>, we store <pid> in the set below.  When the corresponding
+    // program_end event is written, we delete <pid> from the set.  This way, at
+    // lo2s' exit, we know for which processes we are still missing program_end
+    // events.  We then write those missing events, assuming the processes
+    // exited successfully and we simply missed that.
+    std::unordered_set<pid_t> missing_program_end_event_cache_;
 
     std::map<LineInfo, otf2::definition::source_code_location> source_code_locations_;
     std::map<LineInfo, otf2::definition::region> regions_line_info_;
--- a/src/trace/trace.cpp
+++ b/src/trace/trace.cpp
@@ -36,6 +36,9 @@
 #include <nitro/env/get.hpp>
 #include <nitro/env/hostname.hpp>
 
+#include <otf2xx/event/program_begin.hpp>
+#include <otf2xx/event/program_end.hpp>
+
 #include <boost/algorithm/string/replace.hpp>
 #include <boost/format.hpp>
 
@@ -159,7 +162,16 @@ void Trace::begin_record()
 
 void Trace::end_record()
 {
-    stopping_time_ = time::now();
+    // write missing program_end events
+    auto now = lo2s::time::now();
+    for (pid_t pid : missing_program_end_event_cache_)
+    {
+        Log::debug() << "Missed program_begin event for process " << pid
+                     << ", assuming it exited successfully";
+        archive_(thread_sample_locations_.at(pid)) << otf2::event::program_end{ now, EXIT_SUCCESS };
+    }
+
+    stopping_time_ = now;
     Log::info() << "Recording done. Start finalization...";
 }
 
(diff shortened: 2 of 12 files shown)


src/trace/trace.cpp
```diff
@@ -159,7 +162,16 @@ void Trace::begin_record()
 
 void Trace::end_record()
 {
-    stopping_time_ = time::now();
+    // write missing program_end events
+    auto now = lo2s::time::now();
```
bmario: doesn't that technically require the same trick lick here: https://github.com/tud-zih-energy/lo2s/blob/bcf7f78f7c10be11c2c89da31de2075df02fa98e/src/perf/sample/writer.cpp#L212

src/trace/trace.cpp
```diff
@@ -159,7 +162,16 @@ void Trace::begin_record()
 
 void Trace::end_record()
```
bmario: Why is that function here and not part of $INSERT_RANDOM_MONITOR's destructor?

src/trace/trace.cpp
```diff
@@ -281,25 +278,80 @@ otf2::definition::system_tree_node& Trace::intern_process_node(pid_t pid)
...
-    attach_process_location_group(emplaced_node, pid, iname);
+    attach_process_location_group(emplaced_node, pid, prog_name);
+
+    mark_process_begin(pid, prog_name, command_line);
+}
+
+void Trace::mark_process_begin(pid_t pid, const otf2::definition::string& prog_name,
+                               const std::vector<std::string>& command_line)
+{
+    auto now = lo2s::time::now();
```
bmario: same problem like https://github.com/tud-zih-energy/lo2s/blob/bcf7f78f7c10be11c2c89da31de2075df02fa98e/src/perf/sample/writer.cpp#L2112 but the inverse:
How do you guaranty that now is not actually after the first timestamp delivered from perf? Or worse, what happens when the clock sources differ?

src/trace/trace.cpp
```diff
@@ -281,25 +278,80 @@ otf2::definition::system_tree_node& Trace::intern_process_node(pid_t pid)
...
+
+    auto it_proc_loc = thread_sample_locations_.find(pid);
+    if (it_proc_loc == thread_sample_locations_.end())
+    {
+        Log::warn() << "Cannot mark end of unknown process " << pid;
+        return;
+    }
+    const otf2::definition::location& process_location = it_proc_loc->second;
+
+    if (missing_program_end_event_cache_.count(pid) == 0)
```
bmario: use `auto it = missing_program_end_event_cache_.find(pid);` ...

src/trace/trace.cpp
```diff
@@ -281,25 +278,80 @@ otf2::definition::system_tree_node& Trace::intern_process_node(pid_t pid)
...
+    {
+        Log::warn() << "Missing process_begin event for pid " << pid;
+        return;
+    }
+
+    archive_(process_location) << otf2::event::program_end{ now, status_code };
+    missing_program_end_event_cache_.erase(pid);
```
bmario: And then reuse `it` here

src/util.cpp
```diff
@@ -108,6 +108,33 @@ std::string get_process_comm(pid_t pid)
...
+    {
+        std::vector<std::string> cmdline;
+        boost::filesystem::ifstream s{ boost::filesystem::path{ "/proc" } / std::to_string(pid) /
+                                       "cmdline" };
+
+        // /proc/<pid>/cmdline is a NUL-delimited list of arguments
+        for (std::string arg; std::getline(s, arg, '\0');)
+        {
+            Log::debug() << "get_process_cmdline: arg=" << arg;
+            cmdline.emplace_back(std::move(arg));
```
bmario: I don't trust that move... like at all. It scares me. 😨

src/util.cpp
```diff
@@ -108,6 +108,33 @@ std::string get_process_comm(pid_t pid)
...
+std::vector<std::string> get_process_cmdline(pid_t pid)
+{
+    try
+    {
+        std::vector<std::string> cmdline;
+        boost::filesystem::ifstream s{ boost::filesystem::path{ "/proc" } / std::to_string(pid) /
+                                       "cmdline" };
+
+        // /proc/<pid>/cmdline is a NUL-delimited list of arguments
+        for (std::string arg; std::getline(s, arg, '\0');)
```
bmario: in the second iteration, you sort of dereferencing a move-from `arg` here. I'm scared

src/util.cpp
phijor: I concede that this may look frightening, but it should actually be perfectly fine.

`std::getline` calls `arg.erase()` before accessing `arg` in any other way. When moved-from, `std::string`s are in a 'valid, but unspecified state', but `std::string::erase()` has no preconditions, therefore it does not invoke undefined when called on a moved-from object.<sup>[[1]](https://en.cppreference.com/w/cpp/utility/move)</sup>

include/lo2s/trace/trace.hpp
```diff
@@ -114,6 +118,8 @@ class Trace
     void update_process_name(pid_t pid, const std::string& name);
     void update_thread_name(pid_t tid, const std::string& name);
 
+    void mark_process_end(pid_t pid, int status_code);
```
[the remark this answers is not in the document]
phijor: Right now `mark_process_begin` is private, since it's only called internally by `add_process`, while `mark_process_end` needs to be public. So that's the reason they're not grouped together, but I admit that this is not optimal. Note that the public and private interfaces of `update_{process,thread}_name` are not grouped together, too.

Do you have a suggestion how we could group those methods more logically?

src/util.cpp
bmario: Okay, but why not use something in that direction:
```
while(std::getline(s, cmdline.back(),'\0')) {
  // ...
  cmdline.resize(cmdline.size() + 1);
}
```

this would also get rid of the unconventional for-loop-header

phijor: But this will fail on the first iteration, won't it :thinking: Since `cmdline` is empty then and calling `.back()` on it surely is ub?

I'm really not fond of the idea of explicitly modifying `cmdline` (e.g. with `.resize()`) as long as there's `.emplace_back()`. And I don't think we can (or should) eliminate the loop variable. Imo compilers are smart, this function is probably not performance critical and the variable `arg` kind of adds to the expressiveness of this loop.

If your gripe is with the unusual `for`-loop header, one can always do:
```cpp
for (std::string arg;
     std::getline(s, arg, '\0');
     cmdline.emplace_back(std::move(arg)))
    ;
```
which follows the 'initialize-check-update' pattern of the `for`-loop or
```cpp
std::string arg;
while (std::getline(s, arg, '\0')) {
    cmdline.emplace_back(std::move(arg)));
}
```
(though this adds `arg` in the surrounding scope).

bmario: Why I don't like this is because of readability. I'm thinking about some clever combination of back_insterter and std::copy, but I don't think it will directly work this way:

```
std::copy(std::istream_iterator<std::string>(s), std::istream_iterator<std::string>(), std::back_inserter(cmdline));
```

src/trace/trace.cpp
```diff
@@ -159,7 +162,16 @@ void Trace::begin_record()
...
 void Trace::end_record()
 {
-    stopping_time_ = time::now();
+    // write missing program_end events
+    auto now = lo2s::time::now();
+    for (pid_t pid : missing_program_end_event_cache_)
+    {
+        Log::debug() << "Missed program_begin event for process " << pid
+                     << ", assuming it exited successfully";
+        archive_(thread_sample_locations_.at(pid)) << otf2::event::program_end{ now, EXIT_SUCCESS };
```
bmario: Check if this can be moved into the monitor, which holds the referenced location

bmario:
It seems that detached threads are killed together with the main thread. So it should™ be fine to write the end record on the master thread.

tilsche:
Please open a new PR for the remaining stuff in #80

PR #98, opened 2018-10-17
Generate symlink to latest trace

### LINKED ISSUE #95: lo2s-latest trace symlink
Often time one repeatedly runs lo2s and checks if the trace is good. This involves a lot of copying around trace names. A "lo2s-latest" symlink that keeps being update would be really nice.

I suggest making it a toggle argument, but on by default.

Name of the symlink and toggle argument are up for discussion.

cvonelm:
Fixes #95 

If the environment variable LO2S_OUTPUT_LINK is defined, generate a symlink to the latest
trace with $LO2S_OUTPUT_LINK as its name, but only if the path does not exist or is in itself a symlink

### DIFF
--- a/src/trace/trace.cpp
+++ b/src/trace/trace.cpp
@@ -253,22 +253,22 @@ void Trace::create_symlink_to_latest()
 {
     boost::filesystem::path symlink_path = nitro::env::get("LO2S_OUTPUT_LINK");
 
-    if (!symlink_path.empty())
+    if(symlink_path.empty())
     {
-        if (boost::filesystem::is_symlink(symlink_path) || !boost::filesystem::exists(symlink_path))
-        {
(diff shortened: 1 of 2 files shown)


src/trace/trace.cpp
```diff
@@ -244,6 +245,30 @@ Trace::~Trace()
...
     archive_ << system_tree_node_properties_;
+
+    create_symlink_to_latest();
+}
+
+void Trace::create_symlink_to_latest()
+{
+    boost::filesystem::path symlink_path = nitro::env::get("LO2S_OUTPUT_LINK");
+
+    if (!symlink_path.empty())
```
tilsche: prefer early exit.

```suggestion
    if (symlink_path.empty())
    {
        return;
    }
```

PR #99, opened 2018-10-21
Fix typo in README

phijor:


tilsche:
On the bright side it is correct in the paper and on IEEE. See also 22f6f9df713e30da34db14a303798ec95eaef39b

PR #100, opened 2018-10-25
Fix sample writer missing thread begin events

phijor:
For each thread location, thread begin events were written lazily in
conjunction with the first recorded  sample.  If there weren't any
samples recorded for this location, the sample writer would 'forget'
to write the thread begin events.

Now, the sample writer will detect a missing thread begin event before
writing the corresponding thread end event, and write the former with
the earliest timestamp available.

### DIFF
--- a/src/perf/sample/writer.cpp
+++ b/src/perf/sample/writer.cpp
@@ -207,10 +207,32 @@ void Writer::end()
 {
     if (cpuid_ == -1)
     {
-        // time::now() can sometimes can be in the past :-(
-        last_time_point_ = std::max(last_time_point_, lo2s::time::now());
+        // Relative to the timestamps on samples (which stem from a kernel
+        // clock), time::now() can sometimes can be in the past :-(
+        // Therefore, ensure that last_time_point_ >= now.
+        auto now = lo2s::time::now();
+        last_time_point_ = std::max(last_time_point_, now);
+
+        // If we have never written any samples on this location, we also never
+        // got to write the thread_begin event.  Make sure we do that now.
+        if (first_event_)
+        {
+            // first_time_point_ will always be before any other timestamp on
+            // this location.  If no samples were written, it is initialized by
+            // time::now(), which is a monotone clock, therefore it is before
+            // the call to time::now() from above.  If any samples were written,
+            // the required check has occured in handle() above.
+            otf2_writer_ << otf2::event::thread_begin(first_time_point_, trace_.process_comm(pid_),
+                                                      -1);
+        }
+
+        // At this point, transitivity and monotonicity (of lo2s::time::now())
+        // ensure that first_time_point_ <= last_time_point_, therefore samples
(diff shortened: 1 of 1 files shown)


src/perf/sample/writer.cpp
```diff
@@ -208,9 +208,26 @@ void Writer::end()
...
         // time::now() can sometimes can be in the past :-(
-        last_time_point_ = std::max(last_time_point_, lo2s::time::now());
+        auto now = lo2s::time::now();
+        last_time_point_ = std::max(last_time_point_, now);
+
+        // If we have never written any samples on this location, we also never
+        // got to write the thread_begin event.  Make sure we do that now.
+        if (first_event_)
+        {
+            first_time_point_ = std::min(first_time_point_, now);
```
bmario:
```suggestion
```

This check is pointless. `first_time_point_` will always be smaller (or equal -- veeeeeeery unlikely) than `now`. lo2s::time::now() is a steady clock. We discussed that already in the other PR.

src/perf/sample/writer.cpp
```diff
@@ -208,9 +208,26 @@ void Writer::end()
...
+            otf2_writer_ << otf2::event::thread_begin(first_time_point_, trace_.process_comm(pid_),
+                                                      -1);
+
+            // Make sure that our last timepoint is *not* before the first
+            // timepoint.  This way, events on this location span a non-negative
+            // ammount of time.
+            last_time_point_ = std::max(last_time_point_, first_time_point_);
```
bmario:
```suggestion      
```

This check is also not required. A few lines up, we make sure that `last_time_point_ >= now` and `now >= _first_time_point_`. Therefore, here `last_time_point_` is already greater (or equal) to `first_time_point_`.

src/perf/sample/writer.cpp
phijor: *mea culpa.*

src/perf/sample/writer.cpp
phijor: Transitive properties are hard.

src/perf/sample/writer.cpp
```diff
@@ -208,9 +208,26 @@ void Writer::end()
...
+        // got to write the thread_begin event.  Make sure we do that now.
+        if (first_event_)
+        {
+            first_time_point_ = std::min(first_time_point_, now);
+            otf2_writer_ << otf2::event::thread_begin(first_time_point_, trace_.process_comm(pid_),
+                                                      -1);
+
+            // Make sure that our last timepoint is *not* before the first
+            // timepoint.  This way, events on this location span a non-negative
+            // ammount of time.
```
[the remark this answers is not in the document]
phijor: I'm not even sure how this got into *a second* commit.

src/perf/sample/writer.cpp
bmario: I'm confused 😕 It's still in?

PR #106, opened 2018-11-09
Rewrite of the 'exit_reader.hpp' using the PERF_RECORD_COMM event.

### LINKED ISSUE #101: Add an exit reader using perf
Similar to #79 and as suggested in #89, we should add an exit reader based on PERF_RECORD_EXIT

Adding this will allow us to run the system-monitoring mode without root privileges, but with an appropriate paranoid setting.

cvonelm:
Fixes #101.

System monitoring with a perf_event_paranoid setting <= 0 is possible
now.

This uses PERF_RECORD_COMM, as this is the only point where you can get
the process names from inside of perf.


src/monitor/cpu_set_monitor.cpp
```diff
@@ -44,6 +44,8 @@ CpuSetMonitor::CpuSetMonitor() : MainMonitor()
         }
     }
 
+    trace_.add_threads(get_comms_for_running_processes());
```
bmario: Why is it moved up?

cvonelm:
The first comment in https://github.com/tud-zih-energy/lo2s/pull/84 states that it is available on all kernels

src/monitor/cpu_set_monitor.cpp
cvonelm: Because the order in which we do things changed. The old exit_reader would read the name of processes which died during measuring, and then at the end add a processes which are still running. With PERF_RECORD_COMM, we first read the names of the already running processes and then collect the names of all processes that started while measuring. If we would do it like the old exit_reader dit it then we would miss all the processes which just quietly died during measuring

PR #107, opened 2018-11-12
Improve LineInfo for IPs in unresolvable functions

### LINKED ISSUE #102: Rename unresolvable / unknown regions
Users have no clue, what these mean:

- [ ] `?@0x`
- [ ] `?@0xffff`

phijor:
This fixes #102. I'm partial to calling these functions "unresolvable" instead of "unknown". 

----

The source code locations generated for instruction pointers which lie
unresolvable functions are now less cluttered.  Instead of generating
LineInfos (which will later be turned into SCLs and function regions)
that contain the IP, they are now melded together if IP resolution
failed in similar ways:

* IP does not lie in range of a memory map known to lo2s?  LineInfo for
    IP becomes "\<unmapped function\>", located in DSO "\<unknown binary\>".
* IP lies in range of a mapped DSO, but function name resolution fails
    (e.g. for stripped binaries)? LineInfo becomes "\<unknown function\>",
    located in the given DSO.

Including the IP when naming SCLs and function regions only introduces
visual clutter in trace analysis software, as it does not provide any
additionaly information about the sampled function.  Instead, we now
only note that function resolution failed while still indication the
IP's origin on a per-binary basis (if possible).

### DIFF
--- a/include/lo2s/line_info.hpp
+++ b/include/lo2s/line_info.hpp
@@ -32,32 +32,61 @@ namespace lo2s
 {
 struct LineInfo
 {
+private:
+    static const std::string unknown_file_str;
+    static const std::string unknown_binary_str;
+    static const std::string unmapped_function_str;
+    static const std::string unknown_function_str;
+
+    static constexpr unsigned int UNKNOWN_LINE = 0;
+
     // Note: If line is not known, we write 1 anyway so the rest is shown in vampir
-    LineInfo(const std::string& fi, const std::string& fu, unsigned int l, const std::string& d)
-    : file(fi), function(fu), line(l ? l : 1), dso(boost::filesystem::path(d).filename().string())
+    // phijor 2018-11-08: I think this workaround is not needed anymore? vampir
+    // shows source code locations with line == 0 just fine.
+    LineInfo(const std::string& file, const std::string& function, unsigned int line,
+             const std::string& dso)
+    : file(file), function(function), line((line != UNKNOWN_LINE) ? line : 1), dso(dso)
     {
     }
 
-    LineInfo(const char* fi, const char* fu, unsigned int l, const std::string& d)
-    : LineInfo(fi ? std::string(fi) : unknown_str, fu ? std::string(fu) : unknown_str, l, d)
+public:
+    LineInfo() : LineInfo(unknown_file_str, unknown_function_str, UNKNOWN_LINE, unknown_binary_str)
     {
     }
 
-    LineInfo(Address addr) : LineInfo(addr, unknown_str)
+    static LineInfo for_function(const char* file, const char* function, unsigned int line,
+                                 const std::string& dso)
     {
+        return LineInfo((file != nullptr) ? std::string{ file } : unknown_file_str,
+                        (function != nullptr) ? std::string{ function } : unknown_function_str,
+                        line, boost::filesystem::path(dso).filename().string());
     }
 
-    LineInfo(Address addr, const std::string& d)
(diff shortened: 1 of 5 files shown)


include/lo2s/line_info.hpp
```diff
@@ -32,33 +32,62 @@ namespace lo2s
...
+    // phijor 2018-11-08: I think this workaround is not needed anymore? vampir
+    // shows source code locations with line == 0 just fine.
+    LineInfo(const std::string& file, const std::string& function, unsigned int line,
+             const std::string& dso)
+    : file(file), function(function), line((line != UNKNOWN_LINE) ? line : 1), dso(dso)
+    {
+    }
+
+public:
+    LineInfo() : LineInfo(unknown_file_str, unknown_function_str, UNKNOWN_LINE, unknown_binary_str)
```
bmario: If there are "factory" functions for everything, can't we simply delete the default constructor?

include/lo2s/line_info.hpp
```diff
@@ -32,33 +32,62 @@ namespace lo2s
...
+
+    static LineInfo for_function(const char* file, const char* function, unsigned int line,
+                                 const std::string& dso)
+    {
+        return LineInfo((file != nullptr) ? std::string{ file } : unknown_file_str,
+                        (function != nullptr) ? std::string{ function } : unknown_function_str,
+                        line, boost::filesystem::path(dso).filename().string());
+    }
+
+    static LineInfo for_unmapped_function()
```
bmario: What is the difference to `for_unknown_function`?

bmario: Can we use Meyer's singleton?

include/lo2s/line_info.hpp
```diff
@@ -32,33 +32,62 @@ namespace lo2s
 {
 struct LineInfo
 {
+private:
+    static const std::string unknown_file_str;
```
bmario: I'd like to remove these ugly bastards and write the string literals directly at the few places where they are used.

include/lo2s/line_info.hpp
phijor: Sure, I think we don't need that anymore.

include/lo2s/line_info.hpp
phijor: > What is the difference to `for_unknown_function`?

It constructs with `unmapped_function_str` instead of `unknown_function_str`.

> Can we use Meyer's singleton?

Yes, that'd be nice. But right now I don't see how that's more efficient than the current pattern. I'll try to think of something.

bmario: But we only have `unmapped_function/unknown_binary` and `unknown_function/${some binary}`?

bmario: This is what I was thinking:

```suggestion
    static const LineInfo& for_unmapped_function() {
         static LineInfo li(unknown_file_str, unmapped_function_str, UNKNOWN_LINE, unknown_binary_str);
         return li;
    }
```

And then the question really was: Will this be copied from?

phijor: > But we only have `unmapped_function/unknown_binary` and `unknown_function/${some binary}`?

So there's actually [this situation](https://github.com/tud-zih-energy/lo2s/pull/107/files#diff-1080ec2e46093f404fcf651ae4336d2aR594) where we have to "default construct" but I'm not quite sure what a sensible default is.

> And then the question really was: Will this be copied from?
As it stands, yes. `Trace::merge_ips` copies them.

PR #114, opened 2019-03-26
Issue 112 incorrect calling context

### LINKED ISSUE #112: Incorrect calling context
- [ ] Use `tid` instead of `pid` for context switches
 - [ ] Store a separate ip ref map per `tid`
 - [ ] In `merge_calling_contexts`, use a parent of the thread calling context (instead of root)
 - [ ] Check if the `pid` entry can be removed from the ref map (is `tid == pid`?)
 - [ ] For efficiency, cache current ip ref map in the sample writer, but always check if the `sample->tid` equals the current tid (of the last context switch event)
 - [ ] Create small trace with sensible `otf2-print` output
 - [ ] Check if everything is shown correctly in Vampir

cvonelm:
This fixes #112 

This PR adds thread nodes to the calling context tree which serve as parents to the calling context nodes, which previously were directly attached to the trees root

This also fixes an issue in call graph mode (-g) where the calling context reference number was not increased when inner calling context nodes where inserted

[trace_cctx.zip](https://cloudstore.zih.tu-dresden.de/index.php/s/jnoAqmOWoih9r3h) contains a small example trace showcasing the new feature.

### DIFF
--- a/src/perf/sample/writer.cpp
+++ b/src/perf/sample/writer.cpp
@@ -306,7 +306,7 @@ bool Writer::handle(const Reader::RecordCommType* comm)
     {
         summary().register_process(comm->pid);
 
-        comms_[comm->pid] = comm->comm;
+        comms_[comm->tid] = comm->comm;
     }
 
     return false;
--- a/src/perf/sample/writer.cpp
+++ b/src/perf/sample/writer.cpp
@@ -60,8 +60,14 @@ Writer::Writer(pid_t pid, pid_t tid, int cpu, monitor::MainMonitor& Monitor, tra
                                                otf2_writer.location())),
   cpuid_metric_event_(otf2::chrono::genesis(), cpuid_metric_instance_),
   time_converter_(perf::time::Converter::instance()), first_time_point_(lo2s::time::now()),
-  last_time_point_(first_time_point_), next_ip_ref_(0)
+  last_time_point_(first_time_point_)
 {
+    // Must monitor either a CPU or (exclusive) a tid/pid
+    assert((cpu == -1) ^ (pid == -1 && tid == -1));
+    if (tid != -1)
+    {
+        // TODO create empty cctx root for thread monitoring
+    }
 }
 
 Writer::~Writer()
@@ -90,28 +96,24 @@ trace::IpRefMap::iterator Writer::find_ip_child(Address addr, trace::IpRefMap& c
         addr = -2;
     }
     auto ret = children.emplace(std::piecewise_construct, std::forward_as_tuple(addr),
-                                std::forward_as_tuple(next_ip_ref_));
+                                std::forward_as_tuple(next_cctx_ref_));
     if (ret.second)
     {
-        next_ip_ref_++;
+        next_cctx_ref_++;
     }
     return ret.first;
 }
 otf2::definition::calling_context::reference_type
 Writer::cctx_ref(const Reader::RecordSampleType* sample)
 {
-    auto& tid_ip_refs = local_ip_refs_
-                            .emplace(std::piecewise_construct, std::forward_as_tuple(sample->tid),
-                                     std::forward_as_tuple(sample->pid))
-                            .first->second.ip_refs;
     if (!has_cct_)
     {
-        auto it = find_ip_child(sample->ip, tid_ip_refs);
+        auto it = find_ip_child(sample->ip, current_thread_cctx_refs_->second.entry.children);
         return it->second.ref;
     }
     else
     {
-        auto children = &tid_ip_refs;
+        auto children = &current_thread_cctx_refs_->second.entry.children;
         for (uint64_t i = sample->nr - 1;; i--)
         {
             auto it = find_ip_child(sample->ips[i], *children);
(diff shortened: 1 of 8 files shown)


cvonelm:
As evaluating the performance of lo2s was overdue considering that we had some pretty deep changes to  some pretty performance critical paths over the last few PRs, we've agreed to catch up on benchmarking in the last meeting.

My measurements are between lo2s-1.1.1 (pre-refactoring more or less) lo2s-1.2.0 (after the unification of the FdMonitor and the IntervalMonitor) and lo2s-issue112 (state as of this PR), for the process monitoring, system monitoring and "system monitoring with sampling" modes.

For performance evaluation I've used the NPB benchmark (class B), over 6 iterations to eliminate statistical outliers, and at a 2,500,000 sample rate to increase lo2s' impact on performance. (which was more or less the maximum achievable sample rate before i got spammed with perf throttle events) As a performance indicator I've recorded the mop/s per thread value of the NPB Benchmark.

As far as my limited analysis goes, the collected data is sound (the more information we collect in lo2s, the more of an impact it has on performance) and indicates a very slight decrease in lo2s performance as we move forward (in the ballpark of -2 mop/s between the different versions) which shouldn't be too alarming considering that I had +- 5 mop/s differences between different "dry-runs" of the NPB benchmark.

[tests.zip](https://cloudstore.zih.tu-dresden.de/index.php/s/660hoBvWTOCjWlo) contains all files concerning this benchmark:

* report.csv which contains the raw measurements I've collected
* lo2s-${version)-${mode} folders, which are example traces for every combination of lo2s version and mode of operation ("all" stands here as a shorthand for system mode with sampling)
* testing.pl and examples.pl, which are the two perl scripts I've used to automate the measurements and the example trace generation

tilsche:
Remaining tasks

- [ ] Fix <unknown> thread names
- [ ] Create traces with -g

src/trace/trace.cpp
```diff
@@ -677,30 +683,55 @@ void Trace::merge_ips(IpRefMap& new_children, IpCctxMap& children,
...
+        auto tid = local_thread_cctx.first;
+        auto local_ref = local_thread_cctx.second.entry.ref;
+
+        auto global_thread_cctx = calling_context_tree_.find(tid);
+
+        if (global_thread_cctx == calling_context_tree_.end())
+        {
+            if (tid != 0)
+            {
+                // Threads rarely have their own name, in that case use the processes name
```
bmario:
```suggestion
                // Threads rarely have their own name, in that case, use the process' name
```

src/trace/trace.cpp
```diff
@@ -677,30 +683,55 @@ void Trace::merge_ips(IpRefMap& new_children, IpCctxMap& children,
...
+                auto process_cctx = calling_context_tree_.find(local_thread_cctx.second.pid);
+                if(process_cctx == calling_context_tree_.end())
+                {
+                        add_thread(tid, "<unknown>");
```
bmario:
```suggestion
                        add_thread(tid, "<unknown process>");
```

Or unknown thread? 😕

src/perf/sample/writer.cpp
```diff
@@ -120,49 +134,32 @@ Writer::cctx_ref(const Reader::RecordSampleType* sample)
...
-                     << last_time_point_ << ">" << tp;
-        tp = last_time_point_;
-    }
+    tp = adjust_timepoints(tp);
 
     if (sample->pid == 0)
     {
-        last_time_point_ = tp;
+        // WHY???
+        Log::trace() << "sample from pid 0?";
```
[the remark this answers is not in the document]
cvonelm: Yeah, according to [3.4.2.2 of Understanding the Linux Kernel](https://proquest.tech.safaribooksonline.de/book/operating-systems-and-server-administration/linux/0596005652/firstchapter#X2ludGVybmFsX0h0bWxWaWV3P3htbGlkPTA1OTYwMDU2NTIlMkZ1bmRlcnN0YW5kbGtfY2hwXzNfc2VjdF80XzJfMl9odG1sJnF1ZXJ5PXBpZCUyMDA=) (hope this link works) this is the idle process

src/perf/tracepoint/switch_writer.cpp
```diff
@@ -112,16 +115,10 @@ bool SwitchWriter::handle(const Reader::RecordSampleType* sample)
...
-        // If we ever merge interrupt samples and switch events we must use a common counter here!
-        otf2::definition::calling_context::reference_type ref(thread_calling_context_refs_.size());
-        auto ret = thread_calling_context_refs_.emplace(tid, ref);
-        assert(ret.second);
-        it = ret.first;
-    }
-    return it->second;
+    auto ret = thread_calling_context_refs_.emplace(
+        std::piecewise_construct, std::forward_as_tuple(tid),
+        std::forward_as_tuple(-1, thread_calling_context_refs_.size()));
```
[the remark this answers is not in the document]
cvonelm: We do not need a next_cctx_ref_ counter in the tracepoint based switch writer because we only have the top level thread nodes

src/trace/trace.cpp
cvonelm: unknown thread sounds better to me as we are dealing with the tid here

cvonelm:
[this](https://cloudstore.zih.tu-dresden.de/index.php/s/xtmB1v4qTFnKcVt) contains some example traces for this version of lo2s with callgraphs enabled

PR #119, opened 2019-06-17
Add unique_ptr wrappers for C pointers that need to be freed in bfd_resolve

### LINKED ISSUE #115: Fix missing RAII in bfd_resolve.cpp
`handle_` should be a `unique_ptr` with a custom deleter which calls `bfd_close()`. This change will remove so much error spaghetti code.

Also, this line might leak, if `LineInfo::for_function()` throws. Again, use an unique_ptr with custom deleter.
https://github.com/tud-zih-energy/lo2s/blob/b69a467e108dc675e7157c5d48db858c8a160ba2/src/bfd_resolve.cpp#L149-L155

Furthermore, investigate this comment: 
https://github.com/tud-zih-energy/lo2s/blob/b69a467e108dc675e7157c5d48db858c8a160ba2/src/bfd_resolve.cpp#L140

cvonelm:
What the title says, I hope this is good, as I'm not that familiar with modern C++

Fixes #115

### DIFF
--- a/include/lo2s/bfd_resolve.hpp
+++ b/include/lo2s/bfd_resolve.hpp
@@ -107,6 +107,26 @@ private:
     }
 };
 
+struct bfd_handle_deleter
+{
+    void operator()(bfd* p) const
+    {
+        bfd_close(p);
+        p = nullptr;
+    }
+};
+
+struct char_handle_deleter
+{
+    void operator()(char* p) const
+    {
+        free(p);
+    }
+};
+
+using unique_bfd_ptr = std::unique_ptr<bfd, bfd_handle_deleter>;
+using unique_char_ptr = std::unique_ptr<char, char_handle_deleter>;
+
 class Lib
 {
 public:
@@ -115,8 +135,6 @@ public:
     Lib(Lib&&) = delete;
     Lib& operator=(const Lib&) = delete;
     Lib& operator=(Lib&&) = delete;
-
-    ~Lib();
 
     LineInfo lookup(Address addr) const;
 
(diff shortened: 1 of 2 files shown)


include/lo2s/bfd_resolve.hpp
```diff
@@ -107,6 +107,26 @@ class LookupError : public std::runtime_error
     }
 };
 
+struct bfd_handle_deleter
+{
+    void operator()(bfd* p) const
+    {
+        bfd_close(p);
+        p = nullptr;
```
bmario: This assignment only affects the local `p`, which is passed-by-value into this function. So no need for that.

src/bfd_resolve.cpp
```diff
@@ -56,29 +56,25 @@ static boost::filesystem::path check_path(const std::string& name)
...
+    if (!bfd_check_format(handle_.get(), bfd_object))
     {
-        bfd_close(handle_);
-        handle_ = nullptr;
         throw InitError("Not a valid bfd_object", name);
     }
 
     char** matching;
-    if (bfd_check_format_matches(handle_, bfd_archive, &matching))
+    if (bfd_check_format_matches(handle_.get(), bfd_archive, &matching))
```
bmario: According to [this](https://ftp.gnu.org/old-gnu/Manuals/bfd-2.9.1/html_node/bfd_22.html#SEC76), we are leaking memory here. Money quote: 
> When done with the list that `matching` points to, the caller should free it.

As we don't look at the`matching`, why not pass nullptr? Heck, why not just use `bfd_check_format`?

@tilsche

tilsche: I guess we should just use `bfd_check_format` here though I'm not quite sure why we need this check at all...

cvonelm: I've gone ahead and removed  that check all together because we will never end up in that codepath anyways with the other bfd_check_format some lines above

PR #129, opened 2019-08-19
List x86_adapt cpu knobs only if they are not node knobs, also print description

### LINKED ISSUE #125: Improve x86_adapt knob listing
- [ ] If there is the same name for node and cpu knobs, only list the node one (this is used anyway)
- [ ] Show description for knobs

cvonelm:
With this PR, CPU x86_adapt events are only listed if they aren't available as a node event, furthermore print a description for every event

This fixes #125

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,10 +103,10 @@ matrix:
 
 cache:
   directories:
-    - ${TRAVIS_BUILD_DIR}/deps/otf2-2.1
+    - ${TRAVIS_BUILD_DIR}/deps/otf2-2.2
 
 install:
-  - OTF2_VERSION=2.1 # Update cache too when changing this!
+  - OTF2_VERSION=2.2 # Update cache too when changing this!
   - OTF2_DIR=${TRAVIS_BUILD_DIR}/deps/otf2-${OTF2_VERSION}
   - |
     if [ -z "$(ls -A ${OTF2_DIR})" ]; then
--- a/include/lo2s/io.hpp
+++ b/include/lo2s/io.hpp
@@ -36,6 +36,7 @@
 #include <map>
 #include <string>
 
+#include <boost/algorithm/string.hpp>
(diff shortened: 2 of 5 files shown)


tilsche:
@rschoene for some knobs, the description is a very long string. e.g for `Intel_PERFEVTSEL0`

> Performance Event Select Register 0, [0:7] Event Select: Selects a performance event logic unit. [15:8] UMask: Qualifies the microarchitectural condition to detect on the selected event logic. [16] USR: Counts while in privilege level is not ring 0. [17] OS: Counts while in privilege level is ring 0. [18] Edge: Enables edge detection if set. [19] PC: enables pin control. [20] INT: enables interrupt on counter overflow. [21] AnyThread: When set to 1, it enables counting the associated event conditions occurring across all logical processors sharing a processor core. When set to 0, the counter only increments the associated event conditions occurring in the logical processor which programmed the MSR. [22] EN: enables the corresponding performance counter to commence counting when this bit is set. [23] INV: invert the CMASK. [31:24] CMASK: When CMASK is not zero, the corresponding performance counter increments each cycle if the event count is greater than or equal to the CMASK.

It is quite nasty to print that nicely. Would it make sense to introduce newlines before the bit-description `[` in x86_adapt? Then we could indent them consistently at newlines and there would be some structure in it.

rschoene:
@tilsche with https://github.com/tud-zih-energy/x86_adapt/tree/python_knobs_updates you can add \n in the description of knobs. Would that suffice?

rschoene:
I added some \n to knobs. Please have a look at the branch so I can merge it

cvonelm:
Already looks a lot better however some of the newlines are a little bit weird like this one in the middle of a formula:
```
[23:17] Time Window for Power Limit time window for power limit in 2^
[21:17]*(1.0+
[23:22]/10.0) unit according to Intel_RAPL_Power_Unit;
```

PR #130, opened 2019-09-03
Add --perf-readout-interval option

### LINKED ISSUE #127: Make readout interval configurable for watermark monitors
Currently all monitors will be read when a ```--readout-interval``` amount of time has passed. however this is often undesirable for monitors governed by perf watermark, we propose this to be configurable, with the default being that ```--readout-interval``` is only used with monitors that don't support watermarks.

For the command line interface there are two options:
- a ```--watermark-use-readout-interval``` **flag**, which, when on, will lead watermark monitors to use ```--readout-interval```
- a ```--watermark--readout-interval MSEC``` **option** which specifies a readout interval used only with watermark monitors (defaults to 0 which means that no readout interval is used, only watermarks)

cvonelm:
0 means interval based recording is disabled

Fixes #127

### DIFF
--- a/man/lo2s.1.pod
+++ b/man/lo2s.1.pod
@@ -164,11 +164,11 @@ F</proc/sys/kernel/perf_event_mlock_kb>.
 
 =item B<-i>, B<--readout-interval> I<MSEC> (default: C<100>)
 
-Wake up interval based monitors (e.g. x86_adapt, x86_energy) every I<MSEC> milliseconds to read event buffers
+Wake up interval based monitors (i.e. x86_adapt, x86_energy) every I<MSEC> milliseconds to read event buffers
 
 =item B<-I>, B<--perf-readout-interval> I<MSEC> (default: C<0>)
 
-Wake up perf based monitors (e.g. sampling, metrics) at least every I<MSEC> milliseconds to read event buffers.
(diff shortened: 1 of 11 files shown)


man/lo2s.1.pod
```diff
@@ -164,7 +164,12 @@ F</proc/sys/kernel/perf_event_mlock_kb>.
 
 =item B<-i>, B<--readout-interval> I<MSEC> (default: C<100>)
 
-Wake up B<lo2s> at least every I<MSEC> milliseconds to read event buffers.
+Wake up interval based monitors (e.g. x86_adapt, x86_energy) every I<MSEC> milliseconds to read event buffers
```
tilsche: e.g. -> i.e. (check if all are listed)

src/config.cpp
```diff
@@ -226,7 +227,12 @@ void parse_program_options(int argc, const char** argv)
             po::value(&read_interval_ms)
                 ->value_name("MSEC")
                 ->default_value(100),
-            "Maximum amount of time between event buffer readouts in milliseconds.")
+            "Amount of time between readouts of interval based monitors, e.g. x86_adapt, x86_energy.")
```
bmario: Also has `e.g.` instead of `i.e.`

PR #131, opened 2019-09-03
Introduce time interval based metric recording

### LINKED ISSUE #121: Time-interval-based metric recording
Some metrics make sense to record in idle (e.g. residency from sysfs). However, in idle, there is no useful metric-leader for recording metrics.

We should have a fallback to enable recording of metrics with intervals (like previously). Hopefully after the recent monitor refactoring, this isn't a huge deal. We need to discuss the involved monitors (share a monitor with x86_adapt?). Also configuration (confusion with `--metric-frequency`)

cvonelm:
This PR is still somewhat of a request for comments

Time interval based metric recording (TIBMR) is the new default
`--metric-frequency` controls the frequency in TIBMR
Event count based metric recording (ECBMR) is now only used if a `--metric-leader`
is given
`--metric-count` controls the number of events that have to elapse in
ECBMR

`--metric-frequency` can only be used in TIBMR
`--metric-count` can only be used with a `--metric-leader` in ECBMR

Both can  be ommitted as both have default values

When cpu-clock isn't available we try to fallback to the old default metric leaders and print a warning

This fixes #121

### DIFF
--- a/man/lo2s.1.pod
+++ b/man/lo2s.1.pod
@@ -276,32 +276,23 @@ Enable a set of default events for metric recording.
 
 =item B<--metric-leader> I<EVENT>
 
-The leading metric event that controls how often metric events are read from
-shared buffers.
+The leading metric event when using event count based metric recording, use in conjunction with
+B<--metric-count> to control the number of events that have to elapse before a metric read is performed
 
 =item B<--metric-count> I<N>
 
-Instruct B<lo2s> to read metrics from kernel buffers every time I<N> metric
-leader events (see B<--metric-leader>) have occured.
+Controls the number of events that have to elapse before a metric read is performed by the kernel when using event
+count based metric recording with B<--metric-leader>.
 Higher values reduce the overhead incurred by B<lo2s>, but may lead to the
 kernel buffers overflowing, in which case B<lo2s> will miss some metric events.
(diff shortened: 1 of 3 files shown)


tilsche:
@rschoene for some knobs, the description is a very long string. e.g for `Intel_PERFEVTSEL0`

> Performance Event Select Register 0, [0:7] Event Select: Selects a performance event logic unit. [15:8] UMask: Qualifies the microarchitectural condition to detect on the selected event logic. [16] USR: Counts while in privilege level is not ring 0. [17] OS: Counts while in privilege level is ring 0. [18] Edge: Enables edge detection if set. [19] PC: enables pin control. [20] INT: enables interrupt on counter overflow. [21] AnyThread: When set to 1, it enables counting the associated event conditions occurring across all logical processors sharing a processor core. When set to 0, the counter only increments the associated event conditions occurring in the logical processor which programmed the MSR. [22] EN: enables the corresponding performance counter to commence counting when this bit is set. [23] INV: invert the CMASK. [31:24] CMASK: When CMASK is not zero, the corresponding performance counter increments each cycle if the event count is greater than or equal to the CMASK.

It is quite nasty to print that nicely. Would it make sense to introduce newlines before the bit-description `[` in x86_adapt? Then we could indent them consistently at newlines and there would be some structure in it.

PR #140, opened 2019-11-25
Also write Calling Context Enter/Leaves in process_mode

### LINKED ISSUE #135: Process Monitor adds strange things to stack
I use lo2s with `lo2s -q -g bash`, then start 2 processes (bt.W.x and sp.A.x), each with 4 threads. The main thread of each process has bash at the bottom of its stack, then some unknown, then libGOMP, then the executed functions. If I see it correctly, all below libGOMP is broken.

More at `taurus:/trcdata/tracefiles/rschoene/lo2s_issue_135.tgz`

(bash: orange, broken: light green, OMP: brown, sp.A.x: light blue)
lo2s_135

cvonelm:
This results in a more accurate depiction of when which process was active

This Fixes #135

### DIFF
--- a/include/lo2s/perf/event_reader.hpp
+++ b/include/lo2s/perf/event_reader.hpp
@@ -106,6 +106,14 @@ public:
         uint64_t time;
         // struct sample_id sample_id;
     };
+    struct RecordSwitchType
+    {
+        struct perf_event_header header;
+        uint32_t pid, tid;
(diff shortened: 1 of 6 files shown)


bmario:
In regards to https://github.com/tud-zih-energy/lo2s/issues/135#issuecomment-571613472
@rschoene Please try this branch again

PR #144, opened 2020-04-14
Issue 124 trace cpp with registry

### LINKED ISSUE #124: Refactor the lo2s::trace with the usage of otf2::registry
The trace.cpp/hpp is a big mess. We should use the possibilities of the `otf2::registry` to remove lots of it's code:

https://github.com/tud-zih-energy/otf2xx/blob/master/tests/lookup_registry_test.cpp

cvonelm:
This PR bases the functionality of the trace.cpp on the otf2::definition::registry container instead of
using a zoo of home grown std::map()'s.

This might still contain outstanding issues, due to how long this has been in the works but the test traces i did for system and  process mode look normal to me, so I think this is ready for consideration.

This fixes #124

### DIFF
--- a/src/trace/trace.cpp
+++ b/src/trace/trace.cpp
@@ -86,10 +86,9 @@ std::string get_trace_name(std::string prefix = "")
 Trace::Trace()
 : trace_name_(get_trace_name(config().trace_path)), archive_(trace_name_, "traces"),
   registry_(archive_.registry()),
-  interrupt_generator_(0u, intern("perf HW_INSTRUCTIONS"),
-                       otf2::common::interrupt_generator_mode_type::count,
-                       otf2::common::base_type::decimal, 0, config().sampling_period),
-
+  interrupt_generator_(registry_.create<otf2::definition::interrupt_generator>(
+      intern("perf HW_INSTRUCTIONS"), otf2::common::interrupt_generator_mode_type::count,
+      otf2::common::base_type::decimal, 0, config().sampling_period)),
   comm_locations_group_(registry_.create<otf2::definition::comm_locations_group>(
       intern("All pthread locations"), otf2::common::paradigm_type::pthread,
       otf2::common::group_flag_type::none)),
@@ -103,7 +102,6 @@ Trace::Trace()
   system_tree_root_node_(registry_.create<otf2::definition::system_tree_node>(
       intern(nitro::env::hostname()), intern("machine")))
 {
-
     cpuid_metric_class_.add_member(metric_member("CPU", "CPU executing the task",
                                                  otf2::common::metric_mode::absolute_point,
                                                  otf2::common::type::int64, "cpuid"));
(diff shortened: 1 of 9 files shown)


src/trace/trace.cpp
```diff
@@ -845,57 +652,36 @@ void Trace::add_threads(const std::unordered_map<pid_t, std::string>& tid_map)
...
-
-otf2::definition::string Trace::intern(const std::string& name)
+const otf2::definition::string& Trace::intern(const std::string& name)
 {
     std::lock_guard<std::recursive_mutex> guard(mutex_);
 
-    auto ret = strings_.emplace(std::piecewise_construct, std::forward_as_tuple(name),
-                                std::forward_as_tuple(string_ref(), name));
-    return ret.first->second;
+    return registry_.create<otf2::definition::string>(ByString(name), name);
```
bmario, line 684: This can't work: https://github.com/tud-zih-energy/otf2xx/blob/master/include/otf2xx/registry.hpp#L243

However, seeing this, it might be better to rename/add a `(try_)emplace` method to the registry.

cvonelm:
The use of copies over references is fine as far as I'm concerned. (At least that is what i remember from writing this code some months back). Looking at some places where the public member functions with references instead of copies are used, it seems like they all just pass them directly back into the trace class via another public member function

cvonelm:
https://github.com/tud-zih-energy/otf2xx/pull/60 has to be merged before this

src/trace/trace.cpp
```diff
@@ -489,28 +356,30 @@ otf2::writer::local& Trace::cpu_sample_writer(int cpuid)
...
-        std::forward_as_tuple(location_ref(), intern(name), location_groups_cpu_.at(cpuid),
-                              otf2::definition::location::location_type::cpu_thread));
+    // As the cpuid is unique in this context, create only one writer/location per cpuid
+    const auto& location = registry_.emplace<otf2::definition::location>(
+        ByCpuSampleWriter(cpuid), intern(name),
+        registry_.get<otf2::definition::location_group>(ByCpu(cpuid)),
+        otf2::definition::location::location_type::cpu_thread);
 
-    return archive()(location.first->second);
+    comm_locations_group_.add_member(location);
```
bmario, line 365: Where does this line come from? 

Also, add_member is only an [emplace_back](https://github.com/tud-zih-energy/otf2xx/blob/master/include/otf2xx/definition/detail/group_impl.hpp#L122)

src/trace/trace.cpp
```diff
@@ -489,28 +356,30 @@ otf2::writer::local& Trace::cpu_sample_writer(int cpuid)
...
-        std::piecewise_construct, std::forward_as_tuple(cpuid),
-        std::forward_as_tuple(location_ref(), name, location_groups_cpu_.at(cpuid),
-                              otf2::definition::location::location_type::cpu_thread));
+    const auto& location = registry_.emplace<otf2::definition::location>(
+        ByCpuSwitchWriter(cpuid), name,
+        registry_.get<otf2::definition::location_group>(ByCpu(cpuid)),
+        otf2::definition::location::location_type::cpu_thread);
 
-    return archive()(location.first->second);
+    comm_locations_group_.add_member(location);
```
bmario, line 381: Where does this line come from? 

Also, add_member is only an [emplace_back](https://github.com/tud-zih-energy/otf2xx/blob/master/include/otf2xx/definition/detail/group_impl.hpp#L122)

src/trace/trace.cpp
cvonelm: The comm_locations_group_ was previously generated when writing the Archive in `~Trace()`. But as far as im concerned there is nothing wrong with adding them already when creating the writers, as the writer locations don't change

PR #146, opened 2020-05-22
Issue 117 improve list events

### LINKED ISSUE #117: Improve --list-events
The `--list-events` command shall check the vents rather than just listing them.

I'm thinking of something like:

```

  ref-cycles *
  stalled-cycles-backend #
  stalled-cycles-frontend #
  task-clock


(* only available in process-monitoring mode
(# only available in system-monitoring mode
```

We may also want to check if root is required.

cvonelm:
This PR adds an indicator to the list-events switch, which shows if an event is only available in process monitoring or system mode.

I haven't tackled the "maybe also check it root privileges are needed part" but that should be easy to implement as well, as we most probably only have to check for EACCES

This PR resolves #117 

This is Also based on #144, so that should probably be merged first.

### DIFF
--- a/include/lo2s/perf/counter_description.hpp
+++ b/include/lo2s/perf/counter_description.hpp
@@ -34,6 +34,7 @@ namespace perf
 {
 enum class Availability
 {
+    UNAVAILABLE,
     SYSTEM_MODE,
     PROCESS_MODE,
     UNIVERSAL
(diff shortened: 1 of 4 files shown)


include/lo2s/perf/counter_description.hpp
```diff
@@ -32,18 +32,27 @@ namespace lo2s
...
+    UNIVERSAL
+};
+
 struct CounterDescription
 {
     CounterDescription(const std::string& name, perf_type_id type, std::uint64_t config,
                        std::uint64_t config1 = 0)
-    : name(name), type(type), config(config), config1(config1)
+    : name(name), type(type), config(config), config1(config1),
+      availability(Availability::UNIVERSAL)
```
bmario, line 47: The default should be something like `UNAVAILABLE`.

PR #149, opened 2020-06-04
Replace boost::format with fmtlib

cvonelm:
Part 1 of the effort to remove boost from lo2s

This PR replaces boost::format with the venerable fmtlib, as well as removing some boost dependencies that weren't used at all.


cvonelm:
This also seems to remove the cmake complaints from boost.

PR #150, opened 2020-06-04
Issue 147 Part 2: boost::filesystem -> std::filesystem (and C++17)

cvonelm:
This PR get lo2s to C++17 and so replaces boost::filesystem with the all new and shiny std::filesystem

With this I could also remove stuff such as make_iterator_range and boost::fs::ifstream

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,7 @@ set(THREADS_PREFER_PTHREAD_FLAG true)
 find_package(Threads REQUIRED)
 find_package(Doxygen COMPONENTS dot)
 find_package(x86_energy 2.0 CONFIG)
+find_package(StdFilesystem REQUIRED)
 
 CHECK_STRUCT_HAS_BITFIELD("struct perf_event_attr" context_switch linux/perf_event.h HAVE_PERF_RECORD_SWITCH)
 
@@ -186,7 +187,7 @@ target_link_libraries(lo2s
         Boost::program_options
         Binutils::Binutils
         fmt::fmt
-        stdc++fs
+        std::filesystem
 )
 
 # old glibc versions require -lrt for clock_gettime()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,6 +186,7 @@ target_link_libraries(lo2s
         Boost::program_options
         Binutils::Binutils
         fmt::fmt
+        stdc++fs
 )
 
 # old glibc versions require -lrt for clock_gettime()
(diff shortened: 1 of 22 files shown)


CMakeLists.txt
```diff
@@ -181,9 +184,9 @@ target_link_libraries(lo2s
         Nitro::dl
         Threads::Threads
         Boost::program_options
-        Boost::filesystem
         Binutils::Binutils
         fmt::fmt
+        stdc++fs
```
bmario, line 189: That's not gone work. Have a look at https://github.com/metricq/hta 
Also g++-7 has support for filesystem.

cvonelm: https://github.com/metricq/hta/blob/32c5c96b224895568372808bfb38e988e8672cb9/src/storage/file/directory.hpp#L37 if we want to support GCC 7 we would need to include this everywhere, where <filesystem> is included

bmario: No big deal, just use the same header:
https://github.com/metricq/hta/blob/32c5c96b224895568372808bfb38e988e8672cb9/include/hta/filesystem.hpp

tbh. I have no idea why it isn't used in that file you have linked...

tilsche: Do we actually still need this choclate icecream header or can we get rid of a macro by relying on higher versions.
In any case, feel free to include  snarky comment.

bmario: I don't see the reason to drop gcc-7 just to save on a few lines in one header...

cvonelm: I don't know how relevant this is but, as far as I have looked distribution-wise, there does not really seems to be a distribution that ships with 7 that does not have a package for 8 available.

Then again I have only looked into fairly mainstream stuff, I don't know how the situation in HPC is.

bmario: I don't know why we are discussing the support for g++-7. There are virtually no costs, just one header file. You'd need the CMake shit anyways because I bet you that soon the library will be renamed/removed.

tilsche: 1. I don't even know if `experimental/filesystem` as a drop-in replacement is generally correct. And I don't want to think about it (see also  the filesystem-related proposals adopted in GCC 8 https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017)
2. Sooner or later, we're going for C++20 anyway. I would just get rid of that furuncle right now.
3. I mean including a namespace in `std`.

But if we must do it, we should at least add a
```
#pragma GCC warning "The <filesystem> header is not available, trying to use <experimental/filesystem> with a nasty workaround instead. Use at your own risk."
```

PR #153, opened 2020-06-10
Remove the only union from lo2s

### LINKED ISSUE #151: Make clang-format 10.0 formatting consistent with earlier versions.
In a move that could be considered to be quite ingenious for a piece of software whose job it is to ensure a consistent source code style, clang-format has changed the formatting behaviour of the "Allman" brace breaking style.

As we can't expect every developer to have the latest and greatest(tm) clang installed, we have decided to manually recreate the pre-10.0 (?) brace breaking style.

cvonelm:
I think we can live with using 8 byte more memory every time we run lo2s for the sake of more consistent formatting.

This fixes #151 (At least a little)

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -533,6 +533,13 @@ void parse_program_options(int argc, const char** argv)
     // Use time interval based metric recording as a default
     if (!vm.count("metric-leader"))
     {
+        if (metric_frequency == 0)
+        {
+            Log::fatal()
+                << "--metric-frequency should not be zero when using the default metric leader";
+            std::exit(EXIT_FAILURE);
+        }
+
         Log::debug() << "checking if cpu-clock is available...";
         try
         {
@@ -556,6 +563,12 @@ void parse_program_options(int argc, const char** argv)
     }
     else
     {
+        if (metric_count == 0)
+        {
+            Log::fatal() << "--metric-count should not be zero when using a custom metric leader";
+            std::exit(EXIT_FAILURE);
+        }
+
         config.metric_use_frequency = false;
         config.metric_count = metric_count;
     }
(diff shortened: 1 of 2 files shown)


tilsche:
From what I can tell, we do not use the shared data property of the union, so this change should be fine. However, I'd suggest to replace `metric_use_frequency` with
```
metric_use_frequency() {
    assert((metric_frequency != 0) != (metric_count != 0));
    return metric_frequency != 0;
}
```

I assume these can never be sensibly `0`, right?

cvonelm:
I will put in the check for 0.

We set metric_use_frequency in relation to whether we use the default leader (that sensibly supports frequency) or a custom metric leader (for which count(=period) is more sensible).

This check also already checks if both things are set and then complains

src/config.cpp
```diff
@@ -519,6 +519,11 @@ void parse_program_options(int argc, const char** argv)
         config.disassemble = false;
 #endif
     }
+    if (metric_frequency == 0 || metric_count == 0)
```
tilsche, line 522: Hold on, one of them should always be zero, the other nonzero, right?

cvonelm: yeah misinterpreted a line where we initialize them

src/config.cpp
```diff
@@ -533,6 +533,13 @@ void parse_program_options(int argc, const char** argv)
     // Use time interval based metric recording as a default
     if (!vm.count("metric-leader"))
     {
+        if (metric_frequency == 0)
+        {
+            Log::fatal()
+                << "--metric-frequency should not be zero when using the default metric leader";
```
bmario, line 539: I don't think that this either/or selection is the correct behavior.

PR #157, opened 2020-07-15
Receive all process names

### LINKED ISSUE #143: Better support for fork/exec
I'm currently seeing a lot of wrongly mapped pids (mapped to the wrong executable) when using `lo2s mpirun` on a system. Some processes are mapped correctly, some processes are called like their parent process.
I assume that this is due to the mapping being done at the beginning of tracing a process. Depending on whether this is done before or after the the exec call, it gets either the old name and mapping or the new one.
It'd be good if that mapping would be (1) either done after exec (i.e. at the ending of a process) or (2) the mapping could be re-evaluated at an exec (or when it sees that the mapping of memory region changes accordingly).

Version: v1.2.2-40-gf5dbe26
started with: `lo2s -X -e cpu-cycles -E instructions -E cpu-cycles -- mpirun -np 24 bin/lu.C.x`

cvonelm:
This looks more like it:
image

- check for the ptrace EVENT_EXEC event as the name is already updated there
- fix update_process_name (the region.add_member thingy did not work at  all)
- also change the name of the otf2::definition::location

This fixes #143

### DIFF
--- a/include/lo2s/monitor/abstract_process_monitor.hpp
+++ b/include/lo2s/monitor/abstract_process_monitor.hpp
@@ -42,6 +42,8 @@ public:
 
     virtual void exit_process(pid_t pid) = 0;
     virtual void exit_thread(pid_t tid) = 0;
+
+    virtual void update_process_name(pid_t pid, const std::string& name) = 0;
 };
 } // namespace monitor
 } // namespace lo2s
--- a/include/lo2s/monitor/dummy_monitor.hpp
+++ b/include/lo2s/monitor/dummy_monitor.hpp
@@ -61,6 +61,11 @@ public:
     {
         (void)tid;
     }
+
+    virtual void update_process_name([[maybe_unused]] pid_t pid,
+                                     [[maybe_unused]] const std::string& name) override
+    {
+    }
 };
 } // namespace monitor
 } // namespace lo2s
--- a/include/lo2s/monitor/process_monitor.hpp
+++ b/include/lo2s/monitor/process_monitor.hpp
@@ -49,6 +49,8 @@ public:
     void exit_process(pid_t pid) override;
     void exit_thread(pid_t tid) override;
 
+    void update_process_name(pid_t pid, const std::string& name) override;
+
 private:
     std::map<pid_t, ThreadMonitor> threads_;
 };
--- a/src/monitor/process_monitor.cpp
+++ b/src/monitor/process_monitor.cpp
@@ -58,6 +58,11 @@ void ProcessMonitor::insert_thread(pid_t pid, pid_t tid, std::string name, bool 
     trace_.update_thread_name(tid, name);
 }
 
+void ProcessMonitor::update_process_name(pid_t pid, const std::string& name)
+{
+    trace_.update_process_name(pid, name);
+}
+
 void ProcessMonitor::exit_process(pid_t pid)
 {
     exit_thread(pid);
--- a/src/process_controller.cpp
+++ b/src/process_controller.cpp
@@ -161,7 +161,6 @@ void ProcessController::handle_ptrace_event(pid_t child, int event)
         {
             // we need the pid of the new process
             pid_t new_pid = ptrace_geteventmsg(child);
-
             std::string command = get_process_comm(new_pid);
             Log::debug() << "New process " << new_pid << " (" << command << "): forked from "
                          << child;
@@ -217,6 +216,13 @@ void ProcessController::handle_ptrace_event(pid_t child, int event)
                          << e.what();
             throw;
         }
+    }
+    break;
+    case PTRACE_EVENT_EXEC:
+    {
+        std::string name = get_process_comm(child);
+        Log::debug() << "Exec in " << child << " (" << name << ")";
+        monitor_.update_process_name(child, name);
     }
     break;
     case PTRACE_EVENT_EXIT:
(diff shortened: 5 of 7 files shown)


include/lo2s/monitor/dummy_monitor.hpp
```diff
@@ -61,6 +61,11 @@ class DummyMonitor : public AbstractProcessMonitor
     {
         (void)tid;
     }
+
+    virtual void update_process_name(pid_t pid, std::string name) override
+    {
+        (void)pid, (void)name;
```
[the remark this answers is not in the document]
cvonelm: I'm making a separate PR for that because we use (void) at a lot of places

src/process_controller.cpp
```diff
@@ -219,6 +218,14 @@ void ProcessController::handle_ptrace_event(pid_t child, int event)
         }
     }
     break;
+    case PTRACE_EVENT_EXEC:
+    {
+        pid_t new_pid = ptrace_geteventmsg(child);
+        std::string name = get_process_comm(child);
```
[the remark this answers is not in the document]
cvonelm: We don't need this at all, child is already the pid that we want.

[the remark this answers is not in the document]
bmario: I understand the documentation differently:

> If the execing thread is not a thread group leader, the thread ID is reset to thread group leader's ID before this stop. Since Linux 3.0, the former thread ID can be retrieved with PTRACE_GETEVENTMSG. 

If I read this correctly, `child` is only correct for single-thread processes. But if a thread calls exec, this pid will be wrong.

Side note: Using PTRACE_GETEVENTMSG in this case will bump our Linux dependency subtly to 3.0.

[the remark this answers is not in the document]
cvonelm: having exec1.cpp as:
```cpp
#include <iostream>
#include <thread>
#include <unistd.h>
#include <sys/types.h>
#include <errno.h>
#include <string.h>

void work()
{
    std::cerr << "pid: " << getpid() << std::endl;
    std::cerr << "tid: " << gettid() << std::endl;
    const char *name = "exec2";
    execlp(name, name, NULL);
    std::cerr << strerror(errno) << std::endl;
}
int main (int argc, char **argv)
{
    std::thread t1(work);
    t1.join();
    return 0;
}
```
and exec2.cpp as
```cpp
#include <iostream>
#include <sys/types.h>
#include <unistd.h>

int main(int argc, char ** argv)
{
    std::cerr << "exec pid: " << getpid() << std::endl;
    std::cerr << "exec tid: " << gettid() << std::endl;
    return 0;
}
```
Returns something like:

```
pid: 3161
tid: 3162
exec pid: 3161
exex tid: 3161
```

So, as far as I understand it, we don't need the former tid,only the tgid (pid afaik)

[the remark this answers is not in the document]
bmario: Thanks for the code. I see, the important part, which I forogot is that the whole process gets replaced, so the main thread seizes to exists.

include/lo2s/monitor/abstract_process_monitor.hpp
```diff
@@ -42,6 +42,9 @@ class AbstractProcessMonitor
 
     virtual void exit_process(pid_t pid) = 0;
     virtual void exit_thread(pid_t tid) = 0;
+
+    virtual void update_process_name([[maybe_unused]] pid_t pid,
+                                     [[maybe_unused]] const std::string& name) = 0;
```
bmario, line 47: These attributes shouldn't be part of the declaration.

include/lo2s/monitor/process_monitor.hpp
```diff
@@ -49,6 +49,8 @@ class ProcessMonitor : public AbstractProcessMonitor, public MainMonitor
     void exit_process(pid_t pid) override;
     void exit_thread(pid_t tid) override;
 
+    void update_process_name([[maybe_unused]] pid_t pid, [[maybe_unused]] const std::string& name);
```
bmario, line 52: These attributes shouldn't be part of the declaration.

bmario: This should be marked as `override`

src/monitor/process_monitor.cpp
```diff
@@ -58,6 +58,12 @@ void ProcessMonitor::insert_thread(pid_t pid, pid_t tid, std::string name, bool
     trace_.update_thread_name(tid, name);
 }
 
+void ProcessMonitor::update_process_name([[maybe_unused]] pid_t pid,
+                                         [[maybe_unused]] const std::string& name)
```
bmario, line 62: The arguments are used, so the attributes shouldn't be required here?

src/process_controller.cpp
```diff
@@ -219,6 +218,13 @@ void ProcessController::handle_ptrace_event(pid_t child, int event)
         }
     }
     break;
+    case PTRACE_EVENT_EXEC:
+    {
+        std::string name = get_process_comm(child);
```
bmario, line 223: Once the process execs, all other threads of the process are gone, so do we need some cleanup action here?

cvonelm:
```
If the PTRACE_O_TRACEEXIT option is on, PTRACE_EVENT_EXIT will happen before actual  death.   This  applies to [...], and when threads are torn down on execve(2) in a multithreaded process.
```

What we have no should suffice

cvonelm:
The problem with Vampir apparently was, that we had dangling regions_groups at the end, so I just generated them at the end of the trace instead of trying to change the name.

PR #161, opened 2020-08-27
Rework perf_counter.hpp into reader

cvonelm:
The whole perf metrics reading logic was awkwardly split between the counter::Reader and the PerfCounterGroup and adjacent classes (Which I presume predate the Reader), so I've reworked them into the counter::Reader to prepare for the addition of the "simple" counter::Reader.

I've also shifted some stuff around that is clearly only used by the metrics part and nowhere else.


include/lo2s/perf/counter/reader.hpp
```diff
@@ -21,23 +21,75 @@
...
+
+    int fd = perf_try_event_open(&perf_attr, tid, cpuid, group_fd, 0);
+    if (fd < 0)
+    {
+        Log::error() << "perf_event_open for counter failed";
+        throw_errno();
+    }
+    return fd;
+}
+
```
tilsche, line 92: I'd prefer these definitions being in a `.cpp` file. Shouldn't be static.

include/lo2s/perf/counter/reader.hpp
```diff
@@ -46,18 +98,101 @@ class Reader : public EventReader<T>
...
+        leader_attr.type = counter_collection.leader.type;
+        leader_attr.config = counter_collection.leader.config;
+        leader_attr.config1 = counter_collection.leader.config1;
+
+        leader_attr.sample_type = PERF_SAMPLE_TIME | PERF_SAMPLE_READ;
+        leader_attr.freq = config().metric_use_frequency;
+
+        if (leader_attr.freq)
+        {
+            Log::debug() << "PerfCounterGroup: sample_freq: " << config().metric_frequency;
```
tilsche, line 118: Outdated debug output names (3x)

include/lo2s/perf/counter/reader.hpp
```diff
@@ -21,23 +21,75 @@
...
+    if (fd < 0)
+    {
+        Log::error() << "perf_event_open for counter failed";
+        throw_errno();
+    }
+    return fd;
+}
+
 template <class T>
 class Reader : public EventReader<T>
```
tilsche, line 94: Another class in dire need of a documentation header.

include/lo2s/perf/counter/reader.hpp
```diff
@@ -46,18 +98,101 @@ class Reader : public EventReader<T>
...
+            }
+        }
+        ::close(group_leader_fd_);
     }
 
 protected:
-    metric::PerfCounterGroup counters_;
+    int group_leader_fd_;
+    std::vector<int> counters_;
+    CounterBuffer buf_;
```
tilsche, line 193: These two need more descriptive names. How about `counter_fds_` / `counter_buffers_`?

tilsche (changes requested):
The majority of comments are not really related to the refactoring, but I'd like to use the refactoring to improve overall code quality - the biggest issue of which is missing documentation of what the classes do.

I'm not 100% sold in the `event` -> `counter` rename. It's more consistent now, so that's great. But there is a subtle distinction between the terms. The *event* usually specifies what is being observed and the *counter* then exists as a combination of the event specification and the means to actually count these events at runtime. So I guess my core question is whether `CounterDescription`, which has not changed, is aptly named, or whether should be named `EventDescription` - and then name other stuff consistently. I have no strong opinion on that at the moment, so we can discuss it tomorrow.

include/lo2s/perf/counter/counter_buffer.hpp
```diff
@@ -0,0 +1,110 @@
...
+    static constexpr std::size_t header_size()
+    {
+        return sizeof(GroupReadFormat) - sizeof(values);
+    }
+};
+
+class CounterBuffer
+{
+public:
+    CounterBuffer(const CounterBuffer&) = delete;
```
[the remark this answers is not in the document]
cvonelm: There are two different read formats ("simple" and grouped) but if I'm understanding the perf man-page correctly you can also get non-grouped counters in the GroupReadFormat if you specify `perf_attr.read_format = PERF_FORMAT_GROUP`

With that we can sweep the whole "For reasons unbeknownst to man there are two different  read formats" shebang under the rug and just pretend that there is one consistent, beautiful, read format.

cvonelm:
With CounterDescription I'm onboard with changing the name back to EventDescription. its only named that way because I thought that only the perf counter code uses that class

include/lo2s/perf/counter/counter_buffer.hpp
```diff
@@ -46,6 +46,12 @@ struct GroupReadFormat
     }
 };
 
+/* when reading perf counters, the counter value is reported as the amount of events since
+ * the start of measurement, while we want the amount of events since the last read.
+ *
+ * To get the amount of events in regards to last read, this class buffers the last read result
```
tilsche, line 52:
```suggestion
 * To get the amount of events in regards to last read, this class buffers the previous read result
```

include/lo2s/perf/counter/counter_buffer.hpp
```diff
@@ -0,0 +1,110 @@
...
+        }
+        // there is a bug in perf where this is sometimes swapped
+        if (time_enabled > time_running)
+        {
+            return (static_cast<double>(time_enabled) / time_running) * value;
+        }
+        return (static_cast<double>(time_running) / time_enabled) * value;
+    }
+
+    static constexpr std::size_t total_buf_size(std::size_t ncounters)
```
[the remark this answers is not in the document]
cvonelm: The problem here is, that we do not know the number of counters in the GroupReadFormat, and we can't really carry the number of counters as an extra class parameter, because we directly `read()` into the struct, which unless I'm being stupid here could cause problems.

PR #163, opened 2020-09-18
C++17ify lo2s

cvonelm:
While i still think about some stuff, here are some simple changes to use [[maybe_unused]] and std::byte where applicable.

### DIFF
--- a/include/lo2s/perf/event_reader.hpp
+++ b/include/lo2s/perf/event_reader.hpp
@@ -32,6 +32,7 @@
 #include <algorithm>
 #include <cassert>
 #include <cinttypes>
+#include <cstddef>
 #include <cstdint>
 #include <cstring>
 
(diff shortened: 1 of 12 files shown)


tilsche (approved):

bmario (approved):

include/lo2s/perf/event_reader.hpp
```diff
@@ -317,11 +318,11 @@ class EventReader
         return mmap_pages_ * get_page_size();
     }
 
-    char* data()
+    std::byte* data()
     {
         // workaround for old kernels
         // assert(header()->data_offset == get_page_size());
-        return (char*)base + get_page_size();
+        return reinterpret_cast<std::byte*>(base + get_page_size());
```
bmario, line 325:
```suggestion
        return reinterpret_cast<std::byte*>(base) + get_page_size();
```

The parenthesis were wrong.

PR #173, opened 2021-03-02
Implement a Location type to hide a lot of the per-cpu vs per-process stuff

cvonelm:
[This PR in a nutshell](https://www.youtube.com/watch?v=AbSehcT19u0)

On a more serious note: This PR implements a custom Location type to hide a lot of the "Are we in per-cpu or per-process mode" boilerplate code in preparation for the safe counter measuring mode.

There are some smaller shenanigans with thread names currently going on, which I'm still looking into, but besides that this is ready


src/trace/trace.cpp
```diff
@@ -308,91 +321,49 @@ void Trace::add_lo2s_property(const std::string& name, const std::string& value)
...
-        otf2::definition::location::location_type::metric);
-    return archive_(location);
+    return archive_(intern_location);
 }
 
-otf2::writer::local& Trace::named_metric_writer(const std::string& name)
+// Normally we assume that names are unique, i.e. one name one location
+// But there are places where the name comes from an outside source (e.g. plugins)
+// where we can't assume uniqueness
+otf2::writer::local& Trace::non_unique_metric_writer(const std::string& name)
```
bmario, line 362: I don't see the point of the comment here, as this is the only method that takes a name and returns a writer.
Also, I don't like the new name. Maybe `create` or `emplace` as a prefix works better.

bmario (changes requested):
What tests did you perform after the changes yet?

bmario (approved):

PR #175, opened 2021-03-09
feat: add LO2S::STARTING_TIME property

### LINKED ISSUE #170: Add human-readable timestamps to traces
Ideally, there would be one or more new OTF2 records to add the time information and Vampir would allow showing that. 

Until this happens, we can already add a property with a formated string to traces, just like what we are doing with `uname` information.

cvonelm:
Adds a property to the trace that shows a human readable time string of
when measuring started.

This fixes #170

### DIFF
--- a/src/trace/trace.cpp
+++ b/src/trace/trace.cpp
@@ -39,8 +39,8 @@
 
 #include <filesystem>
 
-#include <fmt/core.h>
 #include <fmt/chrono.h>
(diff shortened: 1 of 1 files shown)


bmario (changes requested):
We have fmtlib, use that instead to fix the build issue:

https://fmt.dev/latest/api.html#date-and-time-formatting

PR #193, opened 2021-08-23
fix (recording): Also count lost events from PERF_RECORD_LOST_SAMPLES

cvonelm:
Since Linux 4.3 there is PERF_RECORD_LOST_SAMPLES, which also indicates that events where lost, so record it additionally  to PERF_RECORD_LOST


bmario (changes requested):
Please add CMake or preprocessor magic, so lo2s can still compile on older kernels.

include/lo2s/perf/event_reader.hpp
```diff
@@ -243,6 +251,15 @@ class EventReader
                     Log::info() << "Lost " << lost->lost << " samples during this chunk.";
                     break;
                 }
+#ifdef HAVE_PERF_RECORD_LOST_SAMPLES
+                case PERF_RECORD_LOST_SAMPLES:
+                {
+                    auto lost = (const RecordLostSamplesType*)event_header_p;
+                    lost_samples += lost->lost;
+                    Log::info() << "Lost " << lost->lost << " samples during this chunk.";
```
bmario, line 259: Since the default logging level is already WARN, I think, this should be at least a warning.

include/lo2s/perf/event_reader.hpp
```diff
@@ -243,6 +251,15 @@ class EventReader
                     Log::info() << "Lost " << lost->lost << " samples during this chunk.";
```
bmario, line 251: Since the default logging level is already WARN, I think, this should be at least a warning.

CMakeLists.txt
```diff
@@ -259,6 +259,10 @@ CHECK_NAME_EXISTS(CLOCK_BOOTTIME linux/time.h HAVE_CLOCK_BOOTTIME)
 CHECK_NAME_EXISTS(CLOCK_REALTIME_ALARM linux/time.h HAVE_CLOCK_REALTIME_ALARM)
 CHECK_NAME_EXISTS(CLOCK_BOOTTIME_ALARM linux/time.h HAVE_CLOCK_BOOTTIME_ALARM)
 
+#check if PERF_RECORD_LOST_SAMPLES is available
+
```
bmario, line 263: Random newline is random

bmario (changes requested):

PR #197, opened 2021-12-14
feat(block I/O): Implement Block I/O tracing

### LINKED ISSUE #196: block I/O tracing
I'm separating this from #194 so we can have a nice high level discussion there and get into the nitty gritty details of the implementation here.

# event reading

by design there is one perf event stream per cpu that we read separately. This is problematic in this case, because one thread on one cpu can issue a block I/O request and a completely different thread (usually a kernel thread) that might be on a completely different cpu will receive the completion event.

A small BPF python hack shows, that separate issue/complete CPUs isn't an edge case : The majority of events have different CPUs where they were issued and completed, meaning that just discarding those events isn't an option.

So instead we probably need to cache the issue and complete events at measuring time and try to construct a coherent view based on the local event observations later

cvonelm:
Introduces the --block-io switch to record block I/O events

The implementation is discussed in detail in #196 

This closes #196

### DIFF
--- a/include/lo2s/monitor/bio_monitor.hpp
+++ b/include/lo2s/monitor/bio_monitor.hpp
@@ -1,0 +1,61 @@
+/*
+ * This file is part of the lo2s software.
+ * Linux OTF2 sampling
+ *
+ * Copyright (c) 2017,
+ *    Technische Universitaet Dresden, Germany
+ *
+ * lo2s is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * lo2s is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with lo2s.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <lo2s/perf/bio/event_cacher.hpp>
+#include <lo2s/perf/bio/writer.hpp>
+
+#include <lo2s/monitor/poll_monitor.hpp>
+#include <lo2s/trace/trace.hpp>
+
+#include <map>
+#include <memory>
+
+namespace lo2s
+{
+namespace monitor
+{
+
+class BioMonitor : public PollMonitor
+{
+public:
+    BioMonitor(trace::Trace& trace, Cpu cpu,
+               std::map<dev_t, std::unique_ptr<perf::bio::Writer>>& writers_);
+
+private:
+    void monitor(int fd) override;
+    void initialize_thread() override;
+    void finalize_thread() override;
+
+    std::string group() const override
+    {
+        return "lo2s::BioMonitor";
+    }
+
+private:
+    Cpu cpu_;
+    perf::bio::EventCacher bio_insert_cacher_;
+    perf::bio::EventCacher bio_issue_cacher_;
+    perf::bio::EventCacher bio_complete_cacher_;
+};
+} // namespace monitor
+} // namespace lo2s
--- a/include/lo2s/util.hpp
+++ b/include/lo2s/util.hpp
@@ -22,6 +22,7 @@
 #pragma once
 
 #include <lo2s/execution_scope.hpp>
+#include <lo2s/perf/bio/block_device.hpp>
 #include <lo2s/types.hpp>
 
 #include <filesystem>
@@ -60,49 +61,12 @@ public:
     T& operator[](const std::string& name)
     {
         std::lock_guard<std::mutex> guard(mutex_);
-        if (elements_.count(name) == 0)
-        {
-            elements_.emplace(std::piecewise_construct, std::forward_as_tuple(name),
-                              std::forward_as_tuple(name));
-        }
-        return elements_.at(name);
+        return elements_.try_emplace(name, name).first->second;
     }
 
 private:
     std::unordered_map<std::string, T> elements_;
     std::mutex mutex_;
-};
-
-enum class BlockDeviceType
-{
-    PARTITION,
-    DISK
-};
-struct BlockDevice
-{
-    BlockDevice() : id(0), name(), type(BlockDeviceType::PARTITION), parent(0)
-    {
-    }
-
-    BlockDevice(dev_t id, const std::string& name, BlockDeviceType type, dev_t parent)
-    : id(id), name(name), type(type), parent(parent)
-    {
-    }
-
-    static BlockDevice partition(dev_t id, const std::string& name, dev_t parent)
-    {
-        return BlockDevice(id, name, BlockDeviceType::PARTITION, parent);
-    }
-
-    static BlockDevice disk(dev_t id, std::string name)
-    {
-        return BlockDevice(id, name, BlockDeviceType::DISK, 0);
-    }
-
-    dev_t id;
-    std::string name;
-    BlockDeviceType type;
-    dev_t parent;
 };
 
 std::size_t get_page_size();
@@ -135,5 +99,6 @@ std::unordered_map<Thread, std::string> get_comms_for_running_threads();
 void try_pin_to_scope(ExecutionScope scope);
 
 Thread gettid();
+
 std::vector<BlockDevice> get_block_devices();
 } // namespace lo2s
--- a/include/lo2s/util.hpp
+++ b/include/lo2s/util.hpp
@@ -32,6 +32,7 @@
 #include <mutex>
 #include <string>
 #include <unordered_map>
+#include <vector>
 
 #include <cstdint>
 #include <ctime>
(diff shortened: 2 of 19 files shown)


cvonelm:
TODO: Currently the size of the read/writes is in sectors.

Multiply by 512 byte (unform sector size Linux uses in the generic block layer) to get the real size of the read/writes.

include/lo2s/perf/bio/matcher.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+class Matcher
+{
+public:
+    Matcher(trace::Trace& trace)
+    : trace_(trace), time_converter_(time::Converter::instance()), writer_(trace.bio_writer())
+    {
+    }
+
+    void submit_events(std::vector<BioEvent>& new_events_)
```
bmario, line 82: `new_events_` should be const

man/lo2s.1.pod
```diff
@@ -335,6 +335,20 @@ Record L<x86_energy.h(3)> values.
...
+
+=over
+
+=item B<--block-io>
+
+Record block I/O events. These are based on the block:block_rq_insert for begin events and block:block_rq_complete for end events specifically.
+
+=item B<--block-cache-size> I<NUM>
+
+Size of the per-CPU in number of events. A larger cache size might increase performance but comes at the cost of a higher memor footprint.
```
bmario, line 348:
```suggestion
Size of the per-CPU cache in number-of-events. A larger cache size might increase performance but comes at the cost of a higher memory footprint.
```

man/lo2s.1.pod
```diff
@@ -335,6 +335,20 @@ Record L<x86_energy.h(3)> values.
 
 =back
 
+=head2 B<Block I/O> options
+
+=over
+
+=item B<--block-io>
+
+Record block I/O events. These are based on the block:block_rq_insert for begin events and block:block_rq_complete for end events specifically.
```
bmario, line 344:
```suggestion
Record block I/O events using the block:block_rq_insert tracepoint for begin events and block:block_rq_complete tracepoint for end events.
```

src/config.cpp
```diff
@@ -120,6 +120,7 @@ void parse_program_options(int argc, const char** argv)
     auto& kernel_tracepoint_options = parser.group("Kernel tracepoint options");
     auto& x86_adapt_options = parser.group("x86_adapt options");
     auto& x86_energy_options = parser.group("x86_energy options");
+    auto& io_options = parser.group("I/O tracing options");
```
bmario, line 123:
```suggestion
    auto& io_options = parser.group("I/O recording options");
```

src/monitor/main_monitor.cpp
```diff
@@ -64,6 +64,16 @@ MainMonitor::MainMonitor() : trace_(), metrics_(trace_)
         }
     }
 
+    if (config().use_block_io)
+    {
+        matcher_ = std::make_unique<perf::bio::Matcher>(trace_);
+        for (const auto& cpu : Topology::instance().cpus())
+        {
+            bio_monitors_.emplace_back(std::make_unique<BioMonitor>(trace_, Cpu(cpu.id), matcher_));
```
bmario, line 72: Since C++17, emplace_back returns a `reference`, I'd rather use that to `start()`

src/perf/bio/writer.cpp
```diff
@@ -0,0 +1,55 @@
...
+        mode = otf2::common::io_operation_mode_type::read;
+    }
+    else
+    {
+        mode = otf2::common::io_operation_mode_type::write;
+    }
+    events_.push_back(BioEvent(makedev(record->dev >> 20, record->dev & ((1U << 20) - 1)),
+                               record->sector, sample->time, type_, mode, record->nr_sector));
+
+    if (events_.size() == 1000)
```
bmario, line 45: Shouldn't this compare to `config.block_cache_size` instead?

bmario (changes requested):
I think the class now called `perf::bio::Writer` is a misnomer. It doesn't write events, it caches them. The class called `Matcher` is the actual writer here.

include/lo2s/monitor/bio_monitor.hpp
```diff
@@ -0,0 +1,60 @@
...
+    BioMonitor(trace::Trace& trace, Cpu cpu, std::unique_ptr<perf::bio::Writer>& writer_);
+
+private:
+    void monitor(int fd) override;
+    void initialize_thread() override;
+    void finalize_thread() override;
+
+    std::string group() const override
+    {
+        return "bio::BioMonitor";
```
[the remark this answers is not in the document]
bmario: Or does it add `lo2s::` somewhere else?

cvonelm:
Use Nonblocking instead of blocking becaue it is more accurate to block I/O.
Measure inserts/issues and completes for a more complete view.
Rework exception handling  to better support crashes in reader creation.

bmario:
The generated trace isn't perfect yet. In particular, the system tree is a bit borked.

```
SYSTEM_TREE_NODE                       0  Name: "aldi" <5>, Class: "machine" <4>, Parent: UNDEFINED
[...]
SYSTEM_TREE_NODE                      50  Name: "/dev/sda" <3879>, Class: "block device" <3880>, Parent: "aldi" <0>
SYSTEM_TREE_NODE                      51  Name: "/dev/nvme1n1" <3881>, Class: "block device" <3880>, Parent: "aldi" <0>
[...]
LOCATION_GROUP                        33  Name: "block I/O events" <68>, Type: PROCESS, Parent: "aldi" <0>
LOCATION                               0  Name: "Block I/O events for /dev/sda" <74>, Type: CPU_THREAD, # Events: 27035, Group: "block I/O events" <33>
LOCATION                               1  Name: "Block I/O events for /dev/sda1" <75>, Type: CPU_THREAD, # Events: 0, Group: "block I/O events" <33>
LOCATION                               2  Name: "Block I/O events for /dev/sda2" <76>, Type: CPU_THREAD, # Events: 0, Group: "block I/O events" <33>
LOCATION                               3  Name: "Block I/O events for /dev/nvme1n1" <77>, Type: CPU_THREAD, # Events: 17380, Group: "block I/O events" <33>
LOCATION                               4  Name: "Block I/O events for /dev/nvme1n1p1" <78>, Type: CPU_THREAD, # Events: 0, Group: "block I/O events" <33>
LOCATION                               5  Name: "Block I/O events for /dev/nvme0n1" <79>, Type: CPU_THREAD, # Events: 0, Group: "block I/O events" <33>
LOCATION                               6  Name: "Block I/O events for /dev/nvme0n1p1" <80>, Type: CPU_THREAD, # Events: 0, Group: "block I/O events" <33>
LOCATION                               7  Name: "Block I/O events for /dev/nvme0n1p2" <81>, Type: CPU_THREAD, # Events: 0, Group: "block I/O events" <33>
LOCATION                               8  Name: "Block I/O events for /dev/nvme0n1p3" <82>, Type: CPU_THREAD, # Events: 0, Group: "block I/O events" <33>
LOCATION                               9  Name: "Block I/O events for /dev/nvme0n1p4" <83>, Type: CPU_THREAD, # Events: 0, Group: "block I/O events" <33>
```

- [ ] While there are system tree nodes for "/dev/sda" and "/dev/nvme1n1", there's none for "/dev/nvme0n1"
- [ ] All locations referencing a single location group
- [ ] The only location group directly references the root STN
- [ ] The STNs for the block devices aren't used anywhere

cvonelm:
1. locations are now created on demand too. so there is not going to be a /dev/nvme1n1 location if there aren't any events for it
2. should every block device be its own location group?
3. depends on 2
4. They are used for the creation of IO_REGULAR_FILE's
5. Consistently uses lower case now

bmario:
> 2\. should every block device be its own location group?

Yes.

cvonelm:
Always write system tree node, locations, location_groups for disks, add partitions to location_group.

This information can be retrieved from /sys/dev/block/*/uevent, by using DISKSEQ and DEVTYPE

include/lo2s/util.hpp
```diff
@@ -72,6 +73,37 @@ class StringCache
...
+{
+    PARTITION,
+    DISK
+};
+struct BlockDevice
+{
+    BlockDevice() : id(0), name(), type(BlockDeviceType::PARTITION), parent(0)
+    {
+    }
+    BlockDevice(dev_t id, std::string name, BlockDeviceType type, dev_t parent)
```
bmario, line 86: missing newline between constructors

bmario, line 86:
```suggestion
    BlockDevice(dev_t id, const std::string& name, BlockDeviceType type, dev_t parent)
```

include/lo2s/util.hpp
```diff
@@ -72,6 +73,37 @@ class StringCache
...
+    : id(id), name(name), type(type), parent(parent)
+    {
+    }
+
+    static BlockDevice partition(dev_t id, std::string name, dev_t parent)
```
bmario, line 91:
```suggestion
    static BlockDevice partition(dev_t id, const std::string& name, dev_t parent)
```

include/lo2s/util.hpp
```diff
@@ -72,6 +73,37 @@ class StringCache
...
+    {
+        return BlockDevice(id, name, BlockDeviceType::PARTITION, parent);
+    }
+
+    static BlockDevice disk(dev_t id, std::string name)
```
bmario, line 96:
```suggestion
    static BlockDevice disk(dev_t id, const std::string& name)
```

src/util.cpp
```diff
@@ -247,38 +247,94 @@ Thread gettid()
...
+
+        uint32_t parent_major = 0;
+        uint32_t parent_minor = 0;
+        if (devtype == "partition")
+        {
+            std::filesystem::path parent_dev("/sys");
+
+            // Because someone at Linux has a serious glue-sniffing problem these symlinks are
+            // relative paths and not absolute. Solution: delete the relative part "../../" from the
+            // beginnig and make it absolute
```
bmario, line 307:
```suggestion
            // begining and make it absolute
```

src/util.cpp
```diff
@@ -247,38 +247,94 @@ Thread gettid()
...
+            }
+        }
+
+        uint32_t parent_major = 0;
+        uint32_t parent_minor = 0;
+        if (devtype == "partition")
+        {
+            std::filesystem::path parent_dev("/sys");
+
+            // Because someone at Linux has a serious glue-sniffing problem these symlinks are
```
bmario, line 305: I presume they are relative paths so that someone with an even more serious glue-sniffing problem can (bind) mount `sysfs` to a different path than `/sys`.

bmario (changes requested):

PR #199, opened 2022-01-06
Issue 198 cgroups

### LINKED ISSUE #198: Add CGroup support
With CGroup support (see `--cgroup` flag of `perf record`), one could attach `lo2s` to a subset of all the processes of a system. This would help in Slurm scenarios (where cgroups are generated for Jobs if the plugin is enabled) and for filtering out events from "uninteresting" processes.

cvonelm:
This adds an option to filter perf events by cgroup

@rschoene could you confirm that this does what you want it to do?

Caveats:

* This assumes that cgroupfs is mounted at /sys/fs/cgroup, where it is mounted on 99,9% of Linux systems. However there is nothing stopping you from mounting it wherever you like. Do we:
    1. Live with that
    2. Reguire full paths, which practically no other tool requires
    3. Port perfs rather ugly cgroupfs-mountpoint searching code.
* This only covers perf events, which makes sense for now (why filter x86_energy by cgroup) but we must keep that in mind for some place in the future where not everything in lo2s comes from perf (BPF for example)

This closes #198 .

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -452,36 +452,7 @@ void parse_program_options(int argc, const char** argv)
 
         if (arguments.provided("cgroup"))
         {
-            std::ifstream mtab("/proc/mounts");
-
-            // This regex does not work when cgroupfs is mounted on a path containing whitespaces
-            // But i think people that mount important Linux filesystems on paths containing
-            // whitespaces should not be let anywhere near lo2s anyways
-            // /proc/mounts format:
-            // device mountpoint fs_type options freq passno
-            std::regex cgroup_regex(R"((\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+))");
-            std::smatch cgroup_match;
-            std::string line;
-            while (std::getline(mtab, line))
-            {
-                if (std::regex_match(line, cgroup_match, cgroup_regex))
-                {
-                    // For the ancient cgroupfs mount point we have to use the one with perf_event
-                    // in the options
-                    if (cgroup_match[3].str() != "cgroup" ||
-                        cgroup_match[4].str().find("perf_event"))
-                    {
-                        std::filesystem::path cgroup_path =
-                            std::filesystem::path(cgroup_match[2].str()) / arguments.get("cgroup");
-                        config.cgroup_fd = open(cgroup_path.c_str(), O_RDONLY);
-
-                        if (config.cgroup_fd != -1)
-                        {
-                            break;
-                        }
-                    }
-                }
-            }
+            config.cgroup_fd = get_cgroup_mountpoint_fd(arguments.get("cgroup"));
 
             if (config.cgroup_fd == -1)
             {
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -22,6 +22,7 @@
 #include <lo2s/config.hpp>
 
 #include <lo2s/build_config.hpp>
+
 #include <lo2s/io.hpp>
 #include <lo2s/log.hpp>
 #include <lo2s/perf/event_provider.hpp>
@@ -42,11 +43,14 @@
 
 #include <algorithm>
 #include <cstdlib>
-#include <ctime>   // for CLOCK_* macros
+#include <cstring>
+#include <ctime> // for CLOCK_* macros
+#include <filesystem>
 #include <iomanip> // for std::setw
 
 extern "C"
 {
+#include <fcntl.h>
 #include <unistd.h>
 }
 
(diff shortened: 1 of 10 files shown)


rschoene (changes requested):
Could we use getmntend() for "/proc/mounts" to get the cgroup/perf_event location?

src/config.cpp
```diff
@@ -292,6 +302,7 @@ void parse_program_options(int argc, const char** argv)
     config.process =
         arguments.provided("pid") ? Process(arguments.as<pid_t>("pid")) : Process::invalid();
     config.sampling_event = arguments.get("event");
+
```
bmario, line 305: A wild newline appeared

src/config.cpp
```diff
@@ -430,9 +441,30 @@ void parse_program_options(int argc, const char** argv)
...
+        if (arguments.provided("cgroup"))
+        {
+            std::filesystem::path cgroup_path =
+                std::filesystem::path("/sys/fs/cgroup") / arguments.get("cgroup");
+
+            config.cgroup_fd = open(cgroup_path.c_str(), O_RDONLY);
+            if (config.cgroup_fd == -1)
+            {
+                lo2s::Log::fatal() << "Can not open cgroup path /sys/fs/cgroup/"
```
bmario, line 453:
```suggestion
                Log::fatal() << "Can not open cgroup path /sys/fs/cgroup/"
```

src/config.cpp
```diff
@@ -430,9 +441,30 @@ void parse_program_options(int argc, const char** argv)
...
+                                   << arguments.get("cgroup") << ":" << strerror(errno);
+                std::exit(EXIT_FAILURE);
+            }
+        }
     }
     else
     {
+        if (arguments.provided("cgroup"))
+        {
+            lo2s::Log::fatal()
```
bmario, line 463:
```suggestion
            Log::fatal()
```

bmario (commented):
Looks good, but I'd like to see what the turtle says

cvonelm:
I read the cgroup2 mount point from /etc/mtab now

src/config.cpp
```diff
@@ -175,6 +179,12 @@ void parse_program_options(int argc, const char** argv)
 
     general_options.toggle("list-knobs", "List available x86_adapt CPU knobs.");
 
+    general_options
+        .option("cgroup",
+                "Only record perf events for the given cgroup. Can only be used in system-mode")
+        .metavar("CGROUP")
```
bmario, line 185:
```suggestion
        .metavar("NAME")
```

bmario (changes requested):

cvonelm:
for cgroup (not the second version) we have to use the /proc/mounts entry that has "perf_event" as an  attribute. Implement that code.

src/config.cpp
```diff
@@ -430,9 +440,55 @@ void parse_program_options(int argc, const char** argv)
         {
             config.sampling = true;
         }
+
+        if (arguments.provided("cgroup"))
+        {
+            std::ifstream mtab("/proc/mounts");
```
bmario, line 446: the cgroup parsing should be it's own function

src/config.cpp
```diff
@@ -430,9 +440,55 @@ void parse_program_options(int argc, const char** argv)
...
+                if (std::regex_match(line, cgroup_match, cgroup_regex))
+                {
+                    // For the ancient cgroupfs mount point we have to use the one with perf_event
+                    // in the options
+                    if (cgroup_match[3].str() != "cgroup" ||
+                        cgroup_match[4].str().find("perf_event"))
+                    {
+                        std::filesystem::path cgroup_path =
+                            std::filesystem::path(cgroup_match[2].str()) / arguments.get("cgroup");
+                        config.cgroup_fd = open(cgroup_path.c_str(), O_RDONLY);
```
bmario, line 467: who's closing the fd?

bmario (changes requested):

src/util.cpp
```diff
@@ -244,4 +246,38 @@ Thread gettid()
...
+    std::regex cgroup_regex(R"((\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+))");
+    std::smatch cgroup_match;
+    std::string line;
+    while (std::getline(mtab, line))
+    {
+        if (std::regex_match(line, cgroup_match, cgroup_regex))
+        {
+            // For the ancient cgroupfs mount point we have to use the one with perf_event
+            // in the options
+            if (cgroup_match[3].str() != "cgroup" || cgroup_match[4].str().find("perf_event"))
```
bmario, line 268: Should be more like `match[3] == "cgroup2" || (match[3] == cgroup && match[4] == perf_event)`

PR #201, opened 2022-03-04
feat(usability): Print message on ^C

cvonelm:


src/process_controller.cpp
```diff
@@ -265,6 +265,8 @@ void ProcessController::handle_signal(Thread child, int status)
             // exit if detached from first child (the original sampled process)
             if (child == first_child_)
             {
+                std::cout << "[ lo2s: Child exited. Stopping measurements and closing trace. ]"
+                          << std::endl;
                 Log::info() << "Exiting monitor with status " << 0;
```
bmario, line 270: doube output

bmario (changes requested):

PR #217, opened 2022-08-25
Issue 216 alder lake support

### LINKED ISSUE #216: Support for Alder Lake
See #213 for the some more info on how this issue came to be

Alder lake is hybrid architecture, which consists of powerful P-Cores and energy saving E-Cores.

Our current code assumes  that every core is capable of opening every event, which isn't the case here.

There are generally two types of events available on Alder Lake, if viewed from perf: Events that are available everywhere and Events that are available on just one core type.

## Event available everywhere

Those events basically work like you would expect it from older Intel architectures. If I'm interpreting the documentation correctly, then there are actually no magic "everywhere" events, the kernel is just being smart about opening the right type of event for the right core. The extent of this magic of course remains to be tested, but our Alder Lake system is for reasons of being new and shiny seldomly available for such basic test work.

## Events that are available one only one core type

Some PMU events are only available on one of the two types of cores. Generally, the `cpu_atom` PMU contains the E-Core events and the `cpu_core` PMU contains the P-Core events, but that is not something I would not rely on.

Rather, a new file has been added to the PMU directory which you can get the range of cores with support for the PMU from:

```bash
cvonelm@kallisto:~$ cat /sys/devices/cpu_core/cpus 
0-15
```

Parsing this it should be easy to test if the event is really available by testing if the event can be opened on one of the cores in the range.
In system monitoring readers for that event would only be created for cores in that range.
As far as I know, process monitoring already works out the box, with the caveat that as long as a process is scheduled on a core without support for that event, the event will read 0. Correct pinning is left as an exercise to the user in that case.
`--list-events` would be needed to modified to show core availiblity

cvonelm:
This PR adds lo2s support for the heterogenous Alder Lake architecture.

Sadly, this did not make topdown available as simple perf events as I thought, but all other events work now.

[this archive](https://cloudstore.zih.tu-dresden.de/index.php/s/nQT6YzJ7jL4cEnj) contains a simple example trace recorded with

`lo2s -A -E cpu_core/cache-misses -E cpu_atom/cache-misses`

One of the currently rather ugly things is that I'm creating a `metric_class` for every cpu (As every cpu can possibly have completely different `metric_member`s) which looks like this in Vampir:

image

A fix would be using the same metric_class and writing 0 to events not supported on that core. I think this is a hack we have to take, but I would like prior input. Problems with this approach is that tracking which events are supported and which arent might be costly to do during measurements.

This fixes #216


src/config.cpp
```diff
@@ -76,18 +77,26 @@ static inline void print_availability(std::ostream& os, const std::string& descr
...
         else if (ev.availability == perf::Availability::SYSTEM_MODE)
         {
-            event_names.push_back(ev.name + " #");
+            availability_str = " #";
         }
-        else
+        if (ev.get_supported_cpus() != Topology::instance().cpus())
         {
-            event_names.push_back(ev.name);
+            std::set<Cpu> cpus = ev.get_supported_cpus();
```
bmario, line 93:
```suggestion
            const auto& cpus = ev.get_supported_cpus();
```
Why copy? Also, `auto` is a thing.

src/config.cpp
```diff
@@ -687,10 +697,53 @@ void parse_program_options(int argc, const char** argv)
             std::exit(EXIT_FAILURE);
         }
 
+        try
```
bmario, line 700: This metric setup deserves its own function. Make uncle Bob proud again.

src/config.cpp
```diff
@@ -687,10 +697,53 @@ void parse_program_options(int argc, const char** argv)
...
+    const auto& leader_cpus =
+        perf::EventProvider::get_event_by_name(config.metric_leader).get_supported_cpus();
+
+    if (arguments.given("standard-metrics"))
+    {
+        for (const auto& mem_event : platform::get_mem_events())
+        {
+            config.perf_group_events.emplace_back(mem_event.name);
+        }
+        config.perf_group_events.emplace_back("instructions");
```
bmario, line 724: We have these fancy `EventDescription` class, why do we only store strings in `config.perf_group_events`. I'd bet somewhere else in the code we use a bunch of `get_event_by_name` calls to translate all event names in this thing again.

src/metric/x86_energy/metrics.cpp
```diff
@@ -114,22 +114,22 @@ Metrics::Metrics(trace::Trace& trace)
             case xe::Granularity::SOCKET:
             case xe::Granularity::DIE:
             case xe::Granularity::MODULE:
-                stn = trace.system_tree_package_node(packageid);
+                stn = trace.system_tree_package_node(package);
                 break;
             case xe::Granularity::CORE:
-                stn = trace.system_tree_core_node(coreid, packageid);
+                stn = trace.system_tree_core_node(core);
```
bmario, line 120:
```suggestion
                stn = trace.system_tree_core_node(Topology::instance().get_core_for_cpu(installed_cpu));
```

src/metric/x86_energy/metrics.cpp
```diff
@@ -81,27 +81,27 @@ Metrics::Metrics(trace::Trace& trace)
...
-                auto node = architecture_.get_arch_for_cpu(granularity, cpu.id);
+                auto node = architecture_.get_arch_for_cpu(granularity, cpu.as_int());
                 if (node.id() == index)
                 {
-                    cpuid = cpu.id;
-                    packageid = cpu.package_id;
-                    coreid = cpu.core_id;
+                    cpu = installed_cpu;
+                    core = Topology::instance().get_core_for_cpu(installed_cpu);
+                    package = Topology::instance().get_package_for_cpu(installed_cpu);
```
bmario, line 95:
```suggestion
```

src/metric/x86_energy/metrics.cpp
```diff
@@ -114,22 +114,22 @@ Metrics::Metrics(trace::Trace& trace)
             case xe::Granularity::SOCKET:
             case xe::Granularity::DIE:
             case xe::Granularity::MODULE:
-                stn = trace.system_tree_package_node(packageid);
+                stn = trace.system_tree_package_node(package);
```
bmario, line 117:
```suggestion
                stn = trace.system_tree_package_node(Topology::instance().get_package_for_cpu(installed_cpu));
```

src/metric/x86_energy/metrics.cpp
```diff
@@ -81,27 +81,27 @@ Metrics::Metrics(trace::Trace& trace)
 
         for (auto index = 0; index < architecture_.size(granularity); index++)
         {
-            int cpuid = -1;
-            int packageid = -1;
-            int coreid = -1;
+            Cpu cpu = Cpu::invalid();
+            Package package = Package::invalid();
+            Core core = Core::invalid();
```
bmario, line 86:
```suggestion
```

src/metric/x86_energy/metrics.cpp
```diff
@@ -81,27 +81,27 @@ Metrics::Metrics(trace::Trace& trace)
...
 
-            for (const auto& cpu : lo2s::Topology::instance().cpus())
+            for (const auto& installed_cpu : lo2s::Topology::instance().cpus())
             {
-                auto node = architecture_.get_arch_for_cpu(granularity, cpu.id);
+                auto node = architecture_.get_arch_for_cpu(granularity, cpu.as_int());
```
bmario, line 90:
```suggestion
                auto node = architecture_.get_arch_for_cpu(granularity, installed_cpu.as_int());
```

src/perf/event_provider.cpp
```diff
@@ -469,6 +472,25 @@ const EventDescription sysfs_read_event(const std::string& ev_desc)
     }
     EventDescription event(ev_desc, static_cast<perf_type_id>(type), 0, 0);
 
+    std::ifstream cpus_stream(pmu_path / "cpus");
+    std::string cpus_string;
+    cpus_stream >> cpus_string;
+
+    if (cpus_stream)
+    {
+        const std::regex cpus_regex("([0-9]+)-([0-9]+)");
```
bmario, line 481: I think we need a `parse_int_list` method, as this should technically also be used in the `Topology`.
It should be fine with input like `1,3,5-7` to expand to `int[1,3,5,6,7]`.

src/perf/event_provider.cpp
```diff
@@ -469,6 +472,25 @@ const EventDescription sysfs_read_event(const std::string& ev_desc)
...
+        std::smatch cpus_match;
+        if (std::regex_match(cpus_string, cpus_match, cpus_regex))
+        {
+            std::set<Cpu> cpus;
+            for (int i = stoi(cpus_match[1].str()); i <= stoi(cpus_match[2].str()); i++)
+            {
+                cpus.emplace(Cpu(i));
```
bmario, line 488:
```suggestion
                event.cpus.emplace(Cpu(i));
```

src/perf/event_provider.cpp
```diff
@@ -469,6 +472,25 @@ const EventDescription sysfs_read_event(const std::string& ev_desc)
...
+    std::string cpus_string;
+    cpus_stream >> cpus_string;
+
+    if (cpus_stream)
+    {
+        const std::regex cpus_regex("([0-9]+)-([0-9]+)");
+        std::smatch cpus_match;
+        if (std::regex_match(cpus_string, cpus_match, cpus_regex))
+        {
+            std::set<Cpu> cpus;
```
bmario, line 485:
```suggestion
```

src/perf/event_provider.cpp
```diff
@@ -469,6 +472,25 @@ const EventDescription sysfs_read_event(const std::string& ev_desc)
...
+            for (int i = stoi(cpus_match[1].str()); i <= stoi(cpus_match[2].str()); i++)
+            {
+                cpus.emplace(Cpu(i));
+            }
+            event.cpus = cpus;
```
bmario, line 490:
```suggestion
```

src/perf/event_provider.cpp
bmario: Wait, that even exists: `lo2s::detail::parse_list` in `topology.hpp`

src/trace/trace.cpp
```diff
@@ -119,46 +119,51 @@ Trace::Trace()
     registry_.create<otf2::definition::system_tree_node_domain>(
         system_tree_root_node_, otf2::common::system_tree_node_domain::shared_memory);
     const auto& sys = Topology::instance();
-    for (auto& package : sys.packages())
+    for (auto& cpu : sys.cpus())
```
bmario, line 122: This is where not having the relations between package->core->cpu really hurts, as you constantly need to check whether things are there, making the code hard to understand in comparison. I'm not sure if I like that.

bmario (changes requested):
> One of the currently rather ugly things is that I'm creating a metric_class for every cpu (As every cpu can possibly have completely different metric_members) which looks like this in Vampir:

I think you're overengineering here a bit. Shouldn't all events from the `cpu` PMU be available
on all cpus. And events from `cpu_core` or `cpu_atom` only be available on their distinct groups of cpus? Hence, there is either one `metric_class` for all cpus, or two `metric_classes` with completely distinct CPU groups, which might be a nightmare to track, but should be possible nonetheless.

include/lo2s/perf/counter/group/writer.hpp
```diff
@@ -44,7 +44,7 @@ class Writer : public Reader<Writer>, MetricWriter
     bool handle(const RecordSampleType* sample);
 
 private:
-    const CounterCollection& counters_;
+    const CounterCollection counters_;
```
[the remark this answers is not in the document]
cvonelm: group_counters_of previously was a singleton, so const refs made sense. now the CounterCollection is generated on the fly, so pass by value is more sensible. afaik at least.

src/config.cpp
```diff
@@ -76,18 +77,26 @@ static inline void print_availability(std::ostream& os, const std::string& descr
...
-            event_names.push_back(ev.name + " #");
+            availability_str = " #";
         }
-        else
+        if (ev.get_supported_cpus() != Topology::instance().cpus())
         {
-            event_names.push_back(ev.name);
+            std::set<Cpu> cpus = ev.get_supported_cpus();
+            cpu_str =
+                fmt::format(" [ CPUs {}-{}]", std::min_element(cpus.begin(), cpus.end())->as_int(),
```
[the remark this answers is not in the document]
cvonelm: I've tried cbegin and cend, but that gave me the highest and lowest element in some cases. operator< on Cpu type is correctly implemented, however.

src/perf/counter/counter_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+    else
+    {
+        for (auto& ev : userspace_events_)
+        {
+            if (ev.is_supported_in(scope.scope))
+            {
+                res.counters.emplace_back(ev);
+            }
+        }
+    }
```
bmario, line 146: If the CounterCollection knew its scope, this code could live in that class instead and this function collapses to something like this:

```suggestion
    if (scope.type == MeasurementScopeType::GROUP_METRIC)
    {
        return { scope, group_leader_, group_events_ };
    } else {
        return { scope, userspace__events_ };
    }
```

src/perf/counter/counter_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+        if (group_leader_.is_supported_in(scope))
+        {
+            for (const auto& ev : group_events_)
+            {
+                if (ev.is_supported_in(scope))
+                {
+                    return true;
+                }
+            }
+        }
```
bmario, line 167:
```suggestion
        return group_leader_.is_supported_in(scope) && std::any_of(group_events_.begin(), group_events_.end(), [scope](const auto& ev) { return ev.is_supported_in(scope) });
```

src/perf/counter/counter_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+    else
+    {
+        for (const auto& ev : userspace_events_)
+        {
+            if (ev.is_supported_in(scope))
+            {
+                return true;
+            }
+        }
```
bmario, line 186: Same like above

src/perf/counter/group/reader.cpp
```diff
@@ -49,9 +49,13 @@ namespace group
 
 template <class T>
 Reader<T>::Reader(ExecutionScope scope, bool enable_on_exec)
-: counter_buffer_(requested_group_counters().counters.size() + 1)
+: counter_buffer_(CounterProvider::instance()
```
bmario, line 52: We could trade memory for computation here by making the CounterCollection a member.

bmario: Since we save the CounterCollection in the child class `Writer` anyways, we should do that here instead and only pay once.

src/perf/counter/group/writer.cpp
```diff
@@ -33,7 +33,7 @@ namespace group
 {
 Writer::Writer(ExecutionScope scope, trace::Trace& trace, bool enable_on_exec)
 : Reader(scope, enable_on_exec), MetricWriter(MeasurementScope::group_metric(scope), trace),
-  counters_(requested_group_counters())
+  counters_(CounterProvider::instance().collection_for(MeasurementScope::group_metric(scope)))
```
bmario, line 36: move `counters_` member to base class `Reader`

src/perf/counter/userspace/writer.cpp
```diff
@@ -34,7 +34,7 @@ namespace userspace
 {
 Writer::Writer(ExecutionScope scope, trace::Trace& trace)
 : Reader(scope), MetricWriter(MeasurementScope::userspace_metric(scope), trace),
-  counters_(requested_userspace_counters())
+  counters_(CounterProvider::instance().collection_for(MeasurementScope::userspace_metric(scope)))
```
bmario, line 37: Same thing, move to base class

src/perf/counter/userspace/reader.cpp
```diff
@@ -61,7 +65,9 @@ Reader<T>::Reader(ExecutionScope scope)
 
     timerfd_settime(timer_fd_, TFD_TIMER_ABSTIME, &tspec, NULL);
 
-    for (auto& event : requested_userspace_counters().counters)
+    for (auto& event : CounterProvider::instance()
+                           .collection_for(MeasurementScope::userspace_metric(scope))
+                           .counters)
```
bmario, line 70: use member for CounterCollection

src/perf/counter/userspace/reader.cpp
```diff
@@ -45,8 +45,12 @@ namespace userspace
...
 template <class T>
 Reader<T>::Reader(ExecutionScope scope)
-: counter_buffer_(requested_userspace_counters().counters.size()),
-  data_(requested_userspace_counters().counters.size())
+: counter_buffer_(CounterProvider::instance()
+                      .collection_for(MeasurementScope::userspace_metric(scope))
+                      .counters.size()),
+  data_(CounterProvider::instance()
+            .collection_for(MeasurementScope::userspace_metric(scope))
+            .counters.size())
```
bmario, line 53: use member for CounterCollection

src/perf/event_provider.cpp
```diff
@@ -26,13 +26,16 @@
...
 #include <lo2s/perf/event_provider.hpp>
 #include <lo2s/perf/util.hpp>
+#include <lo2s/topology.hpp>
+#include <lo2s/util.hpp>
 
 #include <filesystem>
 #include <fstream>
 #include <ios>
 #include <limits>
 #include <regex>
```
bmario, line 36:
```suggestion
#include <regex>
#include <set>
```

src/topology.cpp
```diff
@@ -1,57 +1,18 @@
...
 
-        std::ifstream cpu_present(base_path / "present");
-        std::getline(cpu_present, present_list);
-    }
+    std::ifstream cpu_online(base_path / "online");
+    std::getline(cpu_online, online_list);
 
-    auto online = detail::parse_list(online_list);
-    auto present = detail::parse_list(present_list);
+    auto online = parse_list(online_list);
```
bmario, line 15: add another function `parse_list_from_file` which handles the file reading as well

src/perf/event_provider.cpp
```diff
@@ -467,7 +470,19 @@ const EventDescription sysfs_read_event(const std::string& ev_desc)
...
-    EventDescription event(ev_desc, static_cast<perf_type_id>(type), 0, 0);
+
+    std::ifstream cpus_stream(pmu_path / "cpus");
+    std::string cpus_string;
+    cpus_stream >> cpus_string;
+
+    std::set<Cpu> cpus;
+    if (cpus_stream)
+    {
+        auto cpuids = parse_list(cpus_string);
```
bmario, line 481: use `parse_list_from_file`

bmario (changes requested):

include/lo2s/perf/event_description.hpp
```diff
@@ -43,9 +47,10 @@ enum class Availability
 struct EventDescription
 {
     EventDescription(const std::string& name, perf_type_id type, std::uint64_t config,
-                     std::uint64_t config1 = 0, double scale = 1, std::string unit = "#")
+                     std::uint64_t config1 = 0, std::set<Cpu> cpus = std::set<Cpu>(),
+                     double scale = 1, std::string unit = "#")
     : name(name), type(type), config(config), config1(config1), scale(scale), unit(unit),
-      availability(Availability::UNAVAILABLE)
+      availability(Availability::UNAVAILABLE), cpus_(cpus)
```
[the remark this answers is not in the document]
cvonelm: I'm not sure how I would go on to rework event_is openable in `src/perf/event_provider.cpp` in a way so that availibility is passed on EventDescrpiption construction.

include/lo2s/trace/trace.hpp
```diff
@@ -25,7 +25,7 @@
 #include <lo2s/execution_scope.hpp>
 #include <lo2s/line_info.hpp>
 #include <lo2s/mmap.hpp>
-#include <lo2s/perf/counter/counter_collection.hpp>
+#include <lo2s/perf/counter/counter_provider.hpp>
```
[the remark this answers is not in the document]
cvonelm: trace.hpp uses the CounterProvider

include/lo2s/perf/event_description.hpp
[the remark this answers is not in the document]
bmario: I think it's fine for this PR, but please open a new issue so we can track this.

include/lo2s/trace/trace.hpp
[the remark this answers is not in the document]
bmario: It also uses CounterCollection, so it should include that as well.

src/perf/counter/counter_provider.cpp
```diff
@@ -0,0 +1,194 @@
...
+                Log::error() << "Failed to determine a suitable metric leader event";
+                Log::error() << "Try manually specifying one with --metric-leader.";
+
+                throw perf::EventProvider::InvalidEvent(leader);
+            }
+        }
+    }
+    else
+    {
+        group_leader_ = EventProvider::get_event_by_name(leader);
```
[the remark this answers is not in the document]
cvonelm: It gets caught by main with a rather unhelpful error message

```bash
cvonelm@donflamingo ~/projects/lo2s/build (git)-[issue-216-alder-lake-support] % ./lo2s --metric-count 10 --metric-leader bli-bla-blub -A
[7659869934928][pid: 13278][tid: 13278][FATAL]: Aborting: Invalid event: invalid event description format
```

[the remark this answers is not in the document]
cvonelm: improved error message with another try-catch block

src/perf/counter/counter_provider.cpp
cvonelm: I don't know if I'm too dogmatic here, but I personally dont like bringing Scope into CounterCollection as it is designed to be Scope agnostic.

bmario: If you're setting different things from outside of the class, depending on the scope, it's anything but agnostic. You just don't see it in the CounterCollection. I could add this method to `CounterCollection`, which makes the implicit dependency visible. I'd rather have that relation explicit.

```
MeasurementScopeType scope_type() const {
  return group_leader_ ? MeasurementScopeType::GROUP_METRIC : MeasurementScopeType::USER_METRIC;
}

bmario (approved):

PR #225, opened 2022-10-04
Don't use exceptions for control flow

bmario:
It hurts to exit the main loop by exception, when during the stack unwinding another exception gets thrown. We're then directly in std::terminate land and the error message is "Success".

### DIFF
--- a/include/lo2s/process_controller.hpp
+++ b/include/lo2s/process_controller.hpp
@@ -37,6 +37,12 @@ namespace lo2s
 
 class ProcessController
 {
+    enum class SignalHandlingState
+    {
+        KeepRunning,
+        Stop
+    };
+
 public:
     ProcessController(Process child, const std::string& name, bool spawn,
                       monitor::AbstractProcessMonitor& monitor);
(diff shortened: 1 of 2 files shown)


include/lo2s/process_controller.hpp
```diff
@@ -37,6 +37,12 @@ namespace lo2s
 
 class ProcessController
 {
+    enum class SignalHandlingState
+    {
+        keep_running,
```
cvonelm, line 42: Maybe nitpick, but we use uppercase for all other enum members.

src/process_controller.cpp
```diff
@@ -313,7 +316,7 @@ void ProcessController::handle_signal(Thread child, int status)
             summary().set_exit_code(WEXITSTATUS(status));
             throw std::system_error(0, std::system_category());
```
cvonelm, line 317: this still uses an exception for control flow

cvonelm (changes requested):

cvonelm (approved):

PR #229, opened 2022-11-02
Issue 223 generic event caching

### LINKED ISSUE #223: Generic event caching mechanism
lo2s' current approach of directly writing events from the event source to the otf2 trace is becoming a bottleneck as new areas of measurement require combining events from multiple different sources into one coherent, time-ordered stream.

For block I/O, lo2s already contains such a caching mechanism, but it is highly specialized to block events.

As a preperation to process mode syscall recording. We should transform  this specialized caching mechanism into a general mechanism compatible with different events.

Ideally this mechanism would use templating so that an event class with adequate support methods could easily be plugged into this caching mechanism

cvonelm:
This PR introduces a generic MultiReader class, which can be used to combine multiple Readers into one ordered Event Stream and adapts the Block I/O Measurement class to use the MultiReader class.

This fixes #223


cvonelm:
So I've tested this with IOR on charon with the state from master and this branch and could not notice any worse benchmark results or event loss. Although charon seems to have good ol' spinning platters instead of SSDs so I don't think that I have pushed lo2s to the fullest here. (Although like 99% of the requests should end-up in the page-cache anyways so there's that)

bmario (changes requested):

include/lo2s/perf/bio/reader.hpp
```diff
@@ -141,11 +171,24 @@ class Reader : public EventReader<T>
...
+    {
+        return ((RecordBlockSampleType*)get())->time;
+    }
+
+    int fd() const
+    {
+        return fd_;
+    }
+
+    Reader(Reader&& other) : PullReader(std::forward<perf::PullReader>(other)), cpu_(other.cpu_)
```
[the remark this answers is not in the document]
cvonelm: > But why bother implementing it yourself in the first place? I think Reader(Reader&& other) = default; should work just fine.

So I thought, but if the default constructor is used, we end up with a bunch of readers all with the same (invalid) fd_ and lo2s explodes violently.

Why lo2s explodes, I don't know. The move constructor is a holdover from the previous Block I/O code.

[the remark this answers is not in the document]
bmario: Are you talking about the default constructor or the defaulted move constructor?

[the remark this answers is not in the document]
cvonelm: defaulted move constructor, when I use Reader(Reader&& other) = default, it breaks.

src/perf/multi_reader.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include <lo2s/perf/bio/writer.hpp>
+#include <lo2s/perf/multi_reader.hpp>
+
+#include <queue>
+
+namespace lo2s
+{
+namespace perf
+{
+template <class Reader, class ReaderIdentity, class Writer>
```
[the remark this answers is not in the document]
cvonelm: I'm confused, isn't the template... stuff needed before a declaration of a templated function?

src/perf/multi_reader.cpp
```diff
@@ -0,0 +1,94 @@
...
+    auto reader = readers_.emplace(std::piecewise_construct, std::forward_as_tuple(identity),
+                                   std::forward_as_tuple(Reader(identity)));
+    return reader.first->second.fd();
+}
+
+template <class Reader, class ReaderIdentity, class Writer>
+void MultiReader<Reader, ReaderIdentity, Writer>::read()
+{
+    std::priority_queue<ReaderState, std::vector<ReaderState>, std::greater<ReaderState>>
+        earliest_available;
```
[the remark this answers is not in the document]
cvonelm: Why? the content of earliest_available is generated completely new every time read() is called.

src/perf/multi_reader.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    while (!earliest_available.empty())
+    {
+        auto state = earliest_available.top();
+        earliest_available.pop();
+
+        if (readers_.at(state.identity).empty())
+        {
+            continue;
+        }
```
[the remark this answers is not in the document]
cvonelm: I think you misunderstand how the control flow works. If I remove that continue, then the readers, which have no events left, will never be removed from earliest_available and we never get to exit the outer while(!earliest_available.empty()) loop.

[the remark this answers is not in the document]
cvonelm: Nvm im stupid

src/util.cpp
```diff
@@ -341,6 +341,16 @@ std::vector<BlockDevice> get_block_devices()
     return result;
 }
 
+BlockDevice block_device_for(dev_t device)
```
bmario, line 344: This should be a constructor/factory method of `BlockDevice`

bmario (changes requested):

include/lo2s/perf/bio/writer.hpp
```diff
@@ -21,120 +21,83 @@
...
-    void submit_events(const std::vector<BioEvent>& new_events)
+    void write(Reader::IdentityType identity, RecordBlockSampleType* header)
     {
-        std::lock_guard<std::mutex> guard(lock_);
-        for (auto& event : new_events)
+        struct RecordBlockSampleType* event = (RecordBlockSampleType*)header;
+        struct RecordBlock* record = (struct RecordBlock*)&(event->blk);
+
+        otf2::common::io_operation_mode_type mode = otf2::common::io_operation_mode_type::flush;
+        // TODO: Handle the few io operations that arent either reads or write
```
bmario, line 51: Turn the TODO into an issue and remove it from the code

include/lo2s/perf/multi_reader.hpp
```diff
@@ -31,61 +31,64 @@
...
 
     void finalize()
     {
-        for (auto& events : events_)
+        for (auto& reader : readers_)
         {
-            writers_[events.first]->submit_events(events.second);
+            reader.second.stop();
         }
+        read();
```
bmario, line 65: please add a comment, why reading after stop is needed

src/perf/bio/block_device.cpp
```diff
@@ -0,0 +1,137 @@
...
+    return result;
+}
+
+BlockDevice BlockDevice::block_device_for(dev_t device)
+{
+    static std::vector<BlockDevice> devices;
+    if (devices.empty())
+    {
+        auto devices = get_block_devices();
+    }
```
bmario, line 130: Is there a need to check for an empty list multiple times during the runtime? If not, you can use the initializer. It's also thread-safe then.
```suggestion
    static std::vector<BlockDevice> devices = get_block_devices();

```

bmario: Also, why is it a vector, if you're actually using the `dev_t` as the key? Smells like a map to me.

src/perf/multi_reader.cpp
```diff
@@ -0,0 +1,89 @@
...
+    auto reader = readers_.emplace(std::piecewise_construct, std::forward_as_tuple(identity),
+                                   std::forward_as_tuple(Reader(identity)));
+    return reader.first->second.fd();
+}
+
+template <class Reader, class Writer>
+void MultiReader<Reader, Writer>::read()
+{
+    std::priority_queue<ReaderState, std::vector<ReaderState>, std::greater<ReaderState>>
+        earliest_available_;
```
bmario, line 47:
```suggestion
        earliest_available;
```

Don't have local variables with `_`

src/perf/multi_reader.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    while (!earliest_available_.empty())
+    {
+        auto state = earliest_available_.top();
+        earliest_available_.pop();
+
+        while (!readers_.at(state.identity).empty())
+        {
+
+            auto event = std::move(*readers_.at(state.identity).top());
```
bmario, line 65: I don't see the point of dereferencing the pointer here. If it's needed, please explain in a comment.

bmario (changes requested):

src/perf/multi_reader.cpp
cvonelm: After the `readers_.at(state.identity).pop()` a few lines down, the event could be overwritten by the kernel ring buffer, so that why I copy it here.

However, I don't like copying here. So I have changed to logic so that pop() is only called after the event has been processed.

bmario (approved):

PR #249, opened 2023-02-08
build: Link with libsframe and libzstd

bmario:
Newer binutils  (>=2.40) want to be linked against libsframe and libzstd.
To still support older binutils, we guess with this heuristic:
If we find a libsframe, we require a suitable libzstd.

### DIFF
--- a/cmake/FindBinutils.cmake
+++ b/cmake/FindBinutils.cmake
@@ -31,6 +31,8 @@ IfUpdatedUnsetAll(Binutils_USE_STATIC_LIBS
     Z_LIBRARIES
     Libiberty_LIBRARIES
     Bfd_LIBRARIES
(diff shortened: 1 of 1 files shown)


cvonelm:
Building this under Arch Linux with the default options fails due to a missing static libzstd.

However, the build using `lo2s_USE_STATIC_LIBS=OFF` works flawlessly.

As already discussed, it is very unlikely that Arch Linux will start shipping a static libzstd [due to this bug](https://bugs.archlinux.org/task/75018) and their generally static unfriendly policy.

bmario:
We talked about that and came to the conclusion that we want to change the default setting of `lo2s_USE_STATIC_LIBS` from `MOSTLY` to `OFF`. However, that should be a different PR.

I think this is good to go then.

PR #256, opened 2023-03-14
Issue 239 close timerfd

### LINKED ISSUE #239: Timerfd is never closed
`PollMonitor::PollMonitor` creates an fd with `timerfd_create` that is apparently never closed.

cvonelm:
This fixes #239 again.


tilsche (changes requested):
- Why twice?
- Also please move dotr implementation to cpp file.

tilsche (approved):

PR #261, opened 2023-03-28
feat(metrics): Support heterogenous architectures with raw events

cvonelm:
As by their nature, there is no information about which CPUs are supported for any given raw event swimming somewhere in the kernel, the only solution to get the set of supported CPUs is good old trial-and-error

### DIFF
--- a/src/perf/event_provider.cpp
+++ b/src/perf/event_provider.cpp
@@ -413,7 +413,30 @@ static void event_description_update(EventDescription& event, std::uint64_t valu
 const EventDescription raw_read_event(const std::string& ev_desc)
 {
     uint64_t code = std::stoull(ev_desc.substr(1), nullptr, 16);
-    const EventDescription event(ev_desc, PERF_TYPE_RAW, code, 0);
+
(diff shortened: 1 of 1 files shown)


cvonelm:
@tilsche can you verify that this PR has the desired effect? lo2s is able to measure the raw event now, but I can not verify if the measurement results make sense.

src/perf/event_provider.cpp
```diff
@@ -413,7 +413,31 @@ static void event_description_update(EventDescription& event, std::uint64_t valu
 const EventDescription raw_read_event(const std::string& ev_desc)
 {
     uint64_t code = std::stoull(ev_desc.substr(1), nullptr, 16);
-    const EventDescription event(ev_desc, PERF_TYPE_RAW, code, 0);
+
+    const EventDescription test_event(ev_desc, PERF_TYPE_RAW, code, 0);
```
tilsche, line 417: That's now unused, is it?

tilsche (changes requested):
Works for me on the initial problem.

tilsche (approved):

PR #268, opened 2023-04-05
fix(sensors): Clear names for sensors with multiple scopes

### LINKED ISSUE #258: Sensors with multiple scopes
E.g. on kallisto, this thing
```
coretemp-isa-0000
Adapter: ISA adapter
Package id 0:  +27.0°C  (high = +80.0°C, crit = +100.0°C)
Core 0:        +26.0°C  (high = +80.0°C, crit = +100.0°C)
Core 4:        +25.0°C  (high = +80.0°C, crit = +100.0°C)
Core 8:        +23.0°C  (high = +80.0°C, crit = +100.0°C)
Core 12:       +21.0°C  (high = +80.0°C, crit = +100.0°C)
Core 16:       +24.0°C  (high = +80.0°C, crit = +100.0°C)
Core 20:       +20.0°C  (high = +80.0°C, crit = +100.0°C)
Core 24:       +26.0°C  (high = +80.0°C, crit = +100.0°C)
Core 28:       +26.0°C  (high = +80.0°C, crit = +100.0°C)
Core 32:       +26.0°C  (high = +80.0°C, crit = +100.0°C)
Core 33:       +26.0°C  (high = +80.0°C, crit = +100.0°C)
Core 34:       +26.0°C  (high = +80.0°C, crit = +100.0°C)
Core 35:       +26.0°C  (high = +80.0°C, crit = +100.0°C)
Core 36:       +27.0°C  (high = +80.0°C, crit = +100.0°C)
Core 37:       +27.0°C  (high = +80.0°C, crit = +100.0°C)
Core 38:       +26.0°C  (high = +80.0°C, crit = +100.0°C)
Core 39:       +26.0°C  (high = +80.0°C, crit = +100.0°C)
```

Ends up creating multiple metrics all called exactly the same `"coretemp-isa-0000`. The scope should be included in the name to disambiguate.

cvonelm:
Apparently intermixing C strings and C++ strings is a bad idea (tm), so keep it as split as possible by using an intermediate std::vector

This fixes #258

### DIFF
--- a/src/metric/sensors/recorder.cpp
+++ b/src/metric/sensors/recorder.cpp
@@ -87,7 +87,7 @@ Recorder::Recorder(trace::Trace& trace)
 
         std::string chip_name;
         chip_name.resize(chip_name_size);
-        int res = sensors_snprintf_chip_name(chip_name.data(), chip_name.size(), chip);
+        int res = sensors_snprintf_chip_name(chip_name.data(), chip_name.size() + 1, chip);
 
         if (res < 0)
         {


src/metric/sensors/recorder.cpp
```diff
@@ -85,9 +85,10 @@ Recorder::Recorder(trace::Trace& trace)
         int chip_name_size = sensors_snprintf_chip_name(nullptr, 0, chip);
         assert(chip_name_size > 0);
 
-        std::string chip_name;
-        chip_name.resize(chip_name_size);
-        int res = sensors_snprintf_chip_name(chip_name.data(), chip_name.size(), chip);
```
bmario, line 90: There's no need to use `vector`. The patch simply is:

`int res = sensors_snprintf_chip_name(chip_name.data(), chip_name.size() + 1, chip);`

It looks scary, but is valid since C++11, as `std::string::data()` returns a pointer to a null-terminated array, but `std::string::size()` returns the number of characters without counting the null-terminator. Hence, it is fine to add one to `size()`.

src/metric/sensors/recorder.cpp
```diff
@@ -85,9 +85,10 @@ Recorder::Recorder(trace::Trace& trace)
...
+        std::vector<char> chip_name_vec(chip_name_size + 1);
+
+        int res = sensors_snprintf_chip_name(chip_name_vec.data(), chip_name_vec.size(), chip);
+        std::string chip_name(chip_name_vec.data());
```
bmario, line 91: If you still want to use vector, move this line below the check, if `res` is less than 0, so you don't try to construct a string from possibly garbage data.

bmario (changes requested):

bmario (approved):

PR #270, opened 2023-04-06
fix(metrics): Prevent NaNs in sampling from occuring

### LINKED ISSUE #267: Nan collection in counter buffer
I encountered a trace in which a collected metric `uncore_clock/clockticks/` became `NaN` at some point and never reverted to a valid state.

After a brief look at the code, it appears that `CounterBuffer`'s state could become `NaN`  in case `diff_enabled > diff_running` and `diff_running == 0`. We need to avoid those cases. Note that this happend on a Kernel, 5.19.1. Not sure what is special with `time_enabled` / `time_runing` for PMU counters - and not sure if that "swap" bug is still in effect. Maybe we can also simplify the code now and only consider non-broken kernels? In any case there should be logging / handling of cases that result in `NaN`.

cvonelm:
A code path could lead to a NaN if diff_running is 0 as this would result in  a division-by-zero.

This commit solves this problem by just deleting that code path, as there is no evidence that the bug it addresses has existed in recent times, if ever.

This fixes #267


bmario (changes requested):
The commit might fix the issue, but the reasoning and commit message need to be sound. A division-by-zero by itself doesn't produce NaN values. To get a Nan here, `diff_value` must be zero, as a product of plus or minus infinity, and zero would give NaN.

bmario:
I don't want to be that picky, but the commit message could need some touches still. But keep that aside for a moment.

What I'm wondering is this:

- There is forgotten knowledge passed down from the ancients of the SVN galaxy that there might have been a perf bug.
- The bug is: `diff_enabled` and `diff_running` might be switched
- The comment in the current code says: `diff_enabled is always smaller than diff_running`
- This implies that the bug is present if `diff_running` is not smaller than `diff_enabled`
- And presumably, we are going into that code path (otherwise, removing it wouldn't do anything)
- So, we are hitting the bug from the tale of the ancients? Which you claim never happened.

I don't understand it.

Another thing, does the situation occur that `diff_running` is zero, but neither `diff_enabled` nor `diff_value` is?
The current implementation would result in infinity. The current patch would change that to zero. What should be the desired outcome? Why don't we handle the case that `diff_running` is zero in the first if condition? Hence the outcome would be `diff_value`.

cvonelm:
> There is forgotten knowledge passed down from the ancients of the SVN galaxy that there might have been a perf bug.

A bug for which there is zero information of it ever existing, neither in kernel changelogs, Robert and Thomas' knowledge and event the ancient scriptures of the lo2s SVN repository. During trial runs of lo2s, it also did not occur.

> What should be the desired outcome? 

Zero, I think. I don't see a case where a metric that was running 0% of the time during that sample interval should produce something else than `diff_value` of 0 occurences.

tilsche:
Is anything blocking this? I kinda need it :/

cvonelm:
someone should have a final look at the explanatory comment I have included in counter_buffer.hpp to make sure that it doesn't contain obvious bullshit.

Other than that I think we can ship it.

PR #272, opened 2023-04-11
feat(sampling) Remove tracepoints based process switch recording support

cvonelm:
This PR removes support for process switch recording based on tracepoints.

lo2s now requires support for PERF_RECORD_SWITCH, which was added in the 4.3 kernel and is widely backported to older systems. Neither Taurus nor the Levante installation at DKRZ lack PERF_RECORD_SWITCH support.


CMakeLists.txt
```diff
@@ -111,16 +110,12 @@ if(PkgConfig_FOUND)
 endif()
 
 
-CHECK_STRUCT_HAS_BITFIELD("struct perf_event_attr" context_switch linux/perf_event.h HAVE_PERF_RECORD_SWITCH)
```
bmario, line 114: Keep the check and enforce it.

bmario (changes requested):

PR #273, opened 2023-04-11
feat(sampling): Remove support for systems that do not have use_clockid

cvonelm:
use_clockid is present since Linux 4.1 and allows for finer grained control over the used perf clock.

All observed systems, like ZIH's Taurus and DKRZ' Levante, have support for use_clockid, so we can safely remove this codepath.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,8 +115,6 @@ endif()
 CMAKE_DEPENDENT_OPTION(USE_RADARE "Enable Radare support." ON "Radare_FOUND" OFF)
 option(USE_HW_BREAKPOINT_COMPAT "Time synchronization fallback for old kernels without hardware breakpoint support." OFF)
 add_feature_info("USE_HW_BREAKPOINT_COMPAT" USE_HW_BREAKPOINT_COMPAT "Time synchronization fallback for old kernels without hardware breakpoint support.")
-option(USE_PERF_CLOCKID "Enables specifying a custom reference clock for recorded events" ON)
-add_feature_info("USE_PERF_CLOCKID" USE_PERF_CLOCKID "Enables specifying a custom reference clock for recorded events")
 option(IWYU "Developer option for include what you use." OFF)
 option(UML_LOOK "Generate graphs with an UML look" OFF)
 add_feature_info("USE_RADARE" USE_RADARE "Use Radare to add instruction information to samples.")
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -131,7 +129,6 @@ add_feature_info("USE_LIBAUDIT" USE_LIBAUDIT "Use libaudit for syscall name reso
 
 # system configuration checks
 CHECK_INCLUDE_FILES(linux/hw_breakpoint.h HAVE_HW_BREAKPOINT_H)
-CHECK_STRUCT_HAS_MEMBER("struct perf_event_attr" clockid linux/perf_event.h HAVE_PERF_EVENT_ATTR_CLOCKID)
```
bmario, line 134: I'd keep the check and assert that it's true.

CMakeLists.txt
```diff
@@ -326,17 +323,6 @@ else()
     endif()
 endif()
 
-# extra feature for newer kernel version
-if (USE_PERF_CLOCKID)
-    if(NOT HAVE_PERF_EVENT_ATTR_CLOCKID)
-        message(SEND_ERROR "The system is not able to use custom reference clocks for perf events. Consider deactivating USE_PERF_CLOCKID.")
-    endif()
```
bmario, line 333: Keep this

src/config.cpp
```diff
@@ -578,7 +578,7 @@ void parse_program_options(int argc, const char** argv)
         const auto& clock = lo2s::time::ClockProvider::get_clock_by_name(requested_clock_name);
 
         lo2s::Log::debug() << "Using clock \'" << clock.name << "\'.";
-#if defined(USE_PERF_CLOCKID) && !defined(USE_HW_BREAKPOINT_COMPAT)
+#if !defined(USE_HW_BREAKPOINT_COMPAT)
```
bmario, line 581:
```suggestion
#ifndef USE_HW_BREAKPOINT_COMPAT
```

src/perf/util.cpp
```diff
@@ -58,7 +58,7 @@ struct perf_event_attr common_perf_event_attrs()
     attr.size = sizeof(struct perf_event_attr);
     attr.disabled = 1;
 
-#if !defined(USE_HW_BREAKPOINT_COMPAT) && defined(USE_PERF_CLOCKID)
+#if !defined(USE_HW_BREAKPOINT_COMPAT)
```
bmario, line 61:
```suggestion
#ifndef USE_HW_BREAKPOINT_COMPAT
```

src/perf/util.cpp
```diff
@@ -125,7 +125,7 @@ int perf_event_description_open(ExecutionScope scope, const EventDescription& de
     // Needed when scaling multiplexed events, and recognize activation phases
     perf_attr.read_format = PERF_FORMAT_TOTAL_TIME_ENABLED | PERF_FORMAT_TOTAL_TIME_RUNNING;
 
-#if !defined(USE_HW_BREAKPOINT_COMPAT) && defined(USE_PERF_CLOCKID)
+#if !defined(USE_HW_BREAKPOINT_COMPAT)
```
bmario, line 128:
```suggestion
#ifndef USE_HW_BREAKPOINT_COMPAT
```

bmario (changes requested):

PR #276, opened 2023-04-19
fix(EventDescription): Turn EventDescription into real class

### LINKED ISSUE #224: Rework EventDescription to be less C-struct like.
EventDescription is currently a rather ugly plain C structure. Rework to be more C++-y

### LINKED ISSUE #275: Events will be marked as unopenable if exclude_kernel=1
In the code path where lo2s checks if the event is only openable with exclude_kernel=1, the availability is never set, resulting in the event not showing.

cvonelm:
The current EventDescription is a mere struct with exposed struct members etc. This PR attempts to rework the EventDescription into a proper class while integrating some of the Event Handling Code into this new class.

This fixes #224 

This also fixes #275, a bug that prevents events from showing if they could only be opened with exclude_kernel=1

### DIFF
--- a/src/perf/counter/group/reader.cpp
+++ b/src/perf/counter/group/reader.cpp
@@ -57,7 +57,7 @@ Reader<T>::Reader(ExecutionScope scope, bool enable_on_exec)
 
     leader_attr.disabled = 1;
 
-#if !defined(USE_HW_BREAKPOINT_COMPAT) && defined(USE_PERF_CLOCKID)
+#ifndef USE_HW_BREAKPOINT_COMPAT
     leader_attr.use_clockid = config().use_clockid;
     leader_attr.clockid = config().clockid;
 #endif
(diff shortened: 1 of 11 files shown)


bmario:
I'm out of the loop about this. Can you extend the PR description (and commit message) and elaborate on what this PR tries to achieve?

src/perf/counter/group/reader.cpp
```diff
@@ -53,11 +53,19 @@ Reader<T>::Reader(ExecutionScope scope, bool enable_on_exec)
...
+
+#ifndef USE_HW_BREAKPOINT_COMPAT
+    leader_attr.use_clockid = config().use_clockid;
+    leader_attr.clockid = config().clockid;
+#endif
+    // When we poll on the fd given by perf_event_open, wakeup, when our buffer is 80% full
+    // Default behaviour is to wakeup on every event, which is horrible performance wise
+    leader_attr.watermark = 1;
+    leader_attr.wakeup_watermark =
+        static_cast<uint32_t>(0.8 * config().mmap_pages * get_page_size());
```
bmario, line 68: This looks oddly familiar. :spaghetti:

bmario (changes requested):
I really dislike, how this copies perf initialization code everywhere, but only in half of the cases. The other half has a new method as sane replacement.

cvonelm:
Dropping this PR in favour of a more thorough refactoring based on the raii-fd branch.

PR #278, opened 2023-04-27
fix(readme): Bring README in line with legacy support removal

### LINKED ISSUE #246: Deprecate / remove legacy support
I strongly dislike carrying around legacy stuff.

- `#ifdef` macros are annoying
- Refactoring and bug fixes are made more expensive (see #241)
- Vastly increased test cost

Currently we have three legacy crufts:

1. tracepoint switch writer, i.e. `NOT USE_PERF_RECORD_SWITCH`.
While it is relatively encapsulated in it's own source file, it also has 9 `#ifndefs` throughout the code.
Ditching it would technically require Linux kernel 4.3, however, it seems to have been backported in 3.10.0-1127.19.1.el7.x86_64. So there's that.

2. `use_clockid` / `USE_PERF_CLOCKID`.
In itself, it is not quite as invasive with it is mostly used together with 3)
As per the manpage, available since Linux 4.1, but apparently also backported.

3. `hw_breakpoint.h` / `USE_HW_BREAKPOINT_COMPAT`.
Also a very annoying codepath to maintain, but probably no go because it is not only dependent on the kernel version (not sure which), but also on hardware support (not supported on Power9).

cvonelm:
This commit brings the README in line with the recent removal of support for systems without clockid or context switch records.

This bumps up the minimum requirement for Linux to 4.3, although both clockid and context switch records appear to be frequently backported features.

Both --help and the manpage did not need adjustements.

This fixes #246

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,14 +37,14 @@ In both modes, system-level metrics (e.g. tracepoints), are always grouped by th
 
 # Build Requirements
 
- * Linux<sup>1</sup>
+ * Linux (>= 4.3)<sup>1</sup>
  * [OTF2](http://www.vi-hps.org/projects/score-p/index.html) (>= 3.0)
  * libbfd
  * libiberty
  * CMake (>= 3.11)
  * A C++ Compiler with C++17 support and the std::filesystem library (GCC > 7, Clang > 5)
 
-<sup>1</sup>: Use Linux >= 4.1 for best results. Older versions, even the ancient 2.6.32, will work, but with degraded time synchronization.
+<sup>1</sup>: Older kernels can work as the required features are oftentimes backported. Otherwise [lo2s 1.7.0](https://github.com/tud-zih-energy/lo2s/releases/tag/v1.7.0) can be used, which is the newest lo2s version with support for kernels as old as even 2.6.32.
 
 # Optional Build Dependencies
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,12 +60,6 @@ In both modes, system-level metrics (e.g. tracepoints), are always grouped by th
 
    `sudo sysctl kernel.perf_event_paranoid=1`
 
- * Tracepoints, block I/O, syscalls and system-wide monitoring on kernels older than 4.3 requires access to debugfs.
```
bmario, line 65: We still need access to debugfs, just not because of system-monitoring.

README.md
```diff
@@ -37,15 +37,13 @@ In both modes, system-level metrics (e.g. tracepoints), are always grouped by th
...
 
- * Linux<sup>1</sup>
+ * Linux (>= 4.3)
  * [OTF2](http://www.vi-hps.org/projects/score-p/index.html) (>= 3.0)
  * libbfd
  * libiberty
  * CMake (>= 3.11)
  * A C++ Compiler with C++17 support and the std::filesystem library (GCC > 7, Clang > 5)
 
-<sup>1</sup>: Use Linux >= 4.1 for best results. Older versions, even the ancient 2.6.32, will work, but with degraded time synchronization.
```
bmario, line 47: We could mention that lo2s versions before $NEXT_VERSION can be used for older kernels.

As we do breaking changes right now, $NEXT_VERSION will be 2.0, right?

bmario (changes requested):

README.md
cvonelm: Yeah, we should release 2.0 following this.

README.md
cvonelm: This will be addressed in the "improve tracepoint documentation" PR

PR #289, opened 2023-06-01
fix(block-io): Fix block I/O support

### LINKED ISSUE #288: --block-io produces semantically wrong traces
`event->sector` (the location) and `event->nr_sector` (the amount of sectors written) are switched up in the code (where one wants the location of a block I/O access, the `nr_sector` field is used and vice versa), resulting in nonsensical I/O events.

This issue is unrelated to the issue preventing lo2s from correctly capturing the activity of NVMe devices specifically(#290), which is not caused by lo2s incorrectly handling block I/O events, but most likely by NVMe devices (partially) bypassing the block layer.

cvonelm:
nr_sector and sector were switched

in the code path where we need to drop an I/O event if it arrived late, the I/O event was never dropped, resulting in an infinite loop.

This does _not_ fix NVMe support as the cause for that issue is unrelated to our handling of block I/O events.

This fixes #288

### DIFF
--- a/src/perf/multi_reader.cpp
+++ b/src/perf/multi_reader.cpp
@@ -61,7 +61,6 @@ void MultiReader<Reader, Writer>::read()
 
         while (!readers_.at(state.identity).empty())
         {
-            // Need to copy because the even
             auto event = readers_.at(state.identity).top();
             if (event->time > earliest_available.top().time)
             {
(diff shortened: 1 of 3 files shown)


src/perf/multi_reader.cpp
```diff
@@ -73,6 +73,7 @@ void MultiReader<Reader, Writer>::read()
 
             if (event->time < highest_written_)
             {
+                readers_.at(state.identity).pop();
```
bmario, line 76: Add a comment, why we drop the event

bmario (changes requested):
Please add documentation of the fields of `RecordBlock`.

bmario (approved):

PR #292, opened 2023-06-15
fix(perf): use correct atomic memory operations

### LINKED ISSUE #287: Revisit memory ordering when accessing the ring buffer
Currently, we use

```
        std::atomic_thread_fence(std::memory_order_acq_rel);
```

For all head/tail accesses.

- That seems wrong, the write access should probably be guarded differently than the read access. That is the case in `linux/tools/include/linux/ring_buffer.h`, and perf the `perf_event.h` docstring.
- This actually compiles away completely on x86-64 with optimization. Which also seems wrong.

cvonelm:
Use __atomic_load_n and __atomic_store_n. those are equivalent to the the smp_load_acquire and smp_store_release functions used in [ring_buffer.h in the linux kernel](https://github.com/torvalds/linux/blob/master/tools/include/linux/ring_buffer.h).

As a proof, they produce the exact same assembly on 

x86: https://godbolt.org/z/s7cMoe5vK

and 

ARM: https://godbolt.org/z/6oYrKnbzv

This fixes #287

### DIFF
--- a/include/lo2s/perf/event_reader.hpp
+++ b/include/lo2s/perf/event_reader.hpp
@@ -310,26 +310,21 @@ private:
         return (struct perf_event_mmap_page*)base;
     }
 
-    // Regarding the issue of memory barriers:
-    // we simply use the same barriers as perf (see linux/tools/include/linux/ring_buffer.h)
     uint64_t data_head() const
     {
-        auto head = header()->data_head;
-        std::atomic_thread_fence(std::memory_order_acq_rel);
-        return head;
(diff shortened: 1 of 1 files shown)


bmario (changes requested):
The `ring_buffer.h` uses a [read-write memory barrier](https://github.com/torvalds/linux/blob/master/tools/include/linux/ring_buffer.h#LL20C41-L20C50), which I do not see anywhere here.

cvonelm:
The Actual implementation further down uses smp_load_acquire() and smp_store_release()

bmario (approved):

PR #303, opened 2023-09-22
refactor: introduce RAII Fd class

cvonelm:
This introduces a RAII Fd class that automatically closes the Fd if the object gets out of scope.

Additionally, I wrote a WeakFd class for all the places where Fds get passed around without changing ownership.

Valgrind reports 0 leaked Fds


include/lo2s/types.hpp
```diff
@@ -256,6 +257,96 @@ class Package
...
 };
+
+class WeakFd;
+class Fd
+{
+public:
+    explicit Fd(int fd) : fd_(fd)
+    {
+    }
+
```
bmario, line 268: A static method `open()` with the signature of the linux open syscall and `std::optional<Fd>` would be nice.

bmario (commented):
Maybe it's me toying around with to many languages, but `Fd` feels a bit clunky now.

include/lo2s/perf/tracepoint/reader.hpp
```diff
@@ -120,58 +120,28 @@ class Reader : public EventReader<T>
...
-        std::swap(fd_, other.fd_);
-    }
-
-    ~Reader()
-    {
-        if (fd_ != -1)
-        {
-            close(fd_);
-        }
+        fd_ = std::move(other.fd_);
```
bmario, line 139: shouldn't this now be the default implementation?

include/lo2s/types/fd.hpp
```diff
@@ -0,0 +1,120 @@
...
+    friend bool operator==(const WeakFd& lhs, const Fd& rhs);
+    friend bool operator==(const Fd& lhs, const WeakFd& rhs);
+
+    operator WeakFd() const;
+
+private:
+    int fd_;
+};
+
+class WeakFd
```
bmario, line 89: As far as I can see, there is no need for the `WeakFd` class. You should be able to replace it with `const Fd&` and `std::cref()` / `std::reference_wrapper<const Fd>`.

include/lo2s/types/fd.hpp
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    Fd& operator=(Fd&& other)
+    {
+        fd_ = other.fd_;
+        other.fd_ = -1;
+        return *this;
+    }
+
+    int as_int() const
```
bmario, line 58: The `as_int` seems pretty obnoxious in code, maybe define the `int operator*()`?

include/lo2s/types/fd.hpp
```diff
@@ -0,0 +1,120 @@
...
+{
+public:
+    explicit Fd(int fd) : fd_(fd)
+    {
+    }
+
+    friend WeakFd;
+
+    Fd(Fd&) = delete;
+    Fd& operator=(Fd&) = delete;
```
bmario, line 43:
```suggestion
    Fd(const Fd&) = delete;
    Fd& operator=(const Fd&) = delete;
```

include/lo2s/perf/tracepoint/reader.hpp
```diff
@@ -120,58 +120,28 @@ class Reader : public EventReader<T>
...
     }
 
     void stop()
     {
-        auto ret = ioctl(fd_, PERF_EVENT_IOC_DISABLE);
+        auto ret = ioctl(fd_.as_int(), PERF_EVENT_IOC_DISABLE);
```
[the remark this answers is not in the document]
bmario: Same for `fnctl`

src/perf/counter/userspace/.#reader.cpp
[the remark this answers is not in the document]
bmario: Also, whatever this file is, it trips out the poor turtle.

PR #306, opened 2023-10-20
Fix off-by-one in userspace metric writer

cvonelm:
Index 0 in the counter_collection is reserved for the metric leader. However userspace metrics dont have a metric leader, so we have to start counting the events from 1 instead of 0.

### DIFF
--- a/src/perf/counter/userspace/writer.cpp
+++ b/src/perf/counter/userspace/writer.cpp
@@ -52,7 +52,9 @@ bool Writer::handle(std::vector<UserspaceReadFormat>& data)
 
     for (std::size_t i = 0; i < counter_buffer_.size(); i++)
     {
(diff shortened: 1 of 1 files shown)


src/perf/counter/userspace/writer.cpp
```diff
@@ -52,7 +52,7 @@ bool Writer::handle(std::vector<UserspaceReadFormat>& data)
 
     for (std::size_t i = 0; i < counter_buffer_.size(); i++)
     {
-        values[i] = counter_buffer_[i] * counter_collection_.get_scale(i);
+        values[i] = counter_buffer_[i] * counter_collection_.get_scale(i + 1);
```
bmario, line 55: please add a comment, why the +1 is necessary here.

bmario (approved):

PR #314, opened 2024-01-05
Add the ability to have system-mode without process tracking

cvonelm:
This should reduce the overhead in situations where you really only want to track metrics in system-wide mode.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -245,6 +245,12 @@ void parse_program_options(int argc, const char** argv)
     general_options.toggle("list-knobs", "List available x86_adapt CPU knobs.");
 
     system_mode_options
+        .toggle("process-recording", "Record process activity. In system monitoring: "
+                                     "(default: enabled)")
+        .allow_reverse()
+        .default_value(true);
+
+    system_mode_options
         .toggle("all-cpus", "Start in system-monitoring mode for all CPUs. "
                             "Monitor as long as COMMAND is running or until PID exits.")
         .short_name("a");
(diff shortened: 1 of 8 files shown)


src/config.cpp
```diff
@@ -221,9 +221,11 @@ void parse_program_options(int argc, const char** argv)
         .metavar("NAME")
         .optional();
 
+    system_mode_options.toggle("system-wide",
```
bmario, line 224: This option seems confusing. What you want to use is a [reversable toggle](https://github.com/tud-zih-energy/nitro/blob/master/tests/options_test.cpp#L412) called `process-recording` (or something like that) with a default of `true`.

So you'd call lo2s with `.../lo2s -a --no-process-monitoring [...]`.

bmario (changes requested):
Also, update the man page with the new command line option.

include/lo2s/trace/trace.hpp
```diff
@@ -270,6 +270,11 @@ class Trace
                     ByProcess(groups_.get_process(scope.as_thread())))
                 .add_member(intern_location);
         }
+
+        // You can not have a location withouth a writer, so create one here. calling archive_ on
+        // the same location twice is harmless in case some requires a real writer later on
+        archive_(intern_location);
+
```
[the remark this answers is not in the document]
cvonelm: The code is unnecessary confusing here. If metrics are used, a metric instance is created with the recorder location set to the location of the sample_writer (which is what location() creates). With `--system-wide` it can now happen that no sample_writer (and associated otf2-writer) will ever be created as with `--system-wide` it is now possible to disable everything that the sample_writer records, leaving us with a location (created for the metric_instance) which has no otf2-writer, which is something OTF2 really does not like. I alleviate this problem by always creating an otf2-writer for the sample_writer location, even if we dont plan to write something to it.

That being said, the way the location for the sample writer is created is unnecessary complex and I've changed this.

src/config.cpp
cvonelm: One of the grievances I have with the lo2s CLI is that it is not very orthogonal in places.

However here consistency is probably better than trying to shoehorn in orthogonality...

PR #315, opened 2024-02-05
Bump file descriptor rlimit to hard rlimit by default

cvonelm:
The default soft limit for the number of open file descriptors per-process in most Linux systems is 1024. This results in crashes on most HPC systems I've used recently as even simple lo2s invocations will exceed this limit with all the per-core perf_event_open calls.

This microscopic soft limit is in place because select() only allows fd's < 1024. If you do not plan to use select() in your code, it is safe to bump the file descriptor limit from the soft limit to the hard limit.

We need to save and restore the old limit before we start the program under measurement however, as the resource limits are inherited by forked processes and we can not guarantee that the program under measurement does not do stupid stuff with select()

### DIFF
--- a/src/monitor/process_monitor_main.cpp
+++ b/src/monitor/process_monitor_main.cpp
@@ -55,6 +55,9 @@ namespace monitor
 
 [[noreturn]] static void run_command(const std::vector<std::string>& command_and_args)
 {
+    struct rlimit initial_rlimit = initial_rlimit_fd();
+    setrlimit(RLIMIT_NOFILE, &initial_rlimit);
+
     /* kill yourself if the parent dies */
     prctl(PR_SET_PDEATHSIG, SIGHUP);
 
@@ -109,6 +112,7 @@ void process_monitor_main(AbstractProcessMonitor& monitor)
             throw_errno();
         }
     }
+
     if (process == Process::invalid())
     {
         Log::error() << "Fork failed.";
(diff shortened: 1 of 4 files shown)


src/util.cpp
```diff
@@ -366,4 +366,24 @@ std::string get_nec_thread_comm(Thread thread)
     // If no '--' is found, fall back to the complete commandline as a name
     return std::accumulate(args.begin(), args.end(), std::string(""));
 }
+
+struct rlimit save_rlimit_fd()
```
bmario, line 370: Don't call it `save`, if it only saves one time and then is used as getter. Also, which or whom rlimit is saved? Maybe `inherited_rlimit_fd`, `initial_rlimit_fd`, or `parent_process_rlimit_fd` instead?

src/monitor/process_monitor_main.cpp
```diff
@@ -55,6 +55,9 @@ namespace monitor
 
 [[noreturn]] static void run_command(const std::vector<std::string>& command_and_args)
 {
+    struct rlimit saved_rlimit = save_rlimit_fd();
+    setrlimit(RLIMIT_OFILE, &saved_rlimit);
```
bmario, line 59: Why is it `RLIMIT_OFILE` instead of `RLIMIT_NOFILE`?

bmario (approved):

tilsche:
All those calls need error handling.

Issue #294, opened 2023-07-03
CUPTI Support in lo2s


cvonelm:
The CUDA Profiling Tools Interface[1] seems to offer metric event and sampling capabilities that far surpass what is possible with NVML, such as detailed memory events (comparable to what intel offers on x86 CPUs) and a perf like Program Counter sampling interface.

[1] https://docs.nvidia.com/cupti/index.html

cvonelm:
implemented in 975d2d8052aeb70e387b164c71c69cc53fb21174

PR #323, opened 2024-03-11
feat(nvidia): Add cupti support

cvonelm:
This commit adds a --nvidia option, which injects a library into the program under measurement, which records entry and exit into CUDA kernels via CUPTI

We might think about bumping the CMake requirement to 3.24 with this version, as older FindCUDAToolkit.cmake fail to correctly detect CUPTI headers[1].

This implements #294

[1]  https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7608

### DIFF
--- a/include/lo2s/cupti/reader.hpp
+++ b/include/lo2s/cupti/reader.hpp
@@ -1,0 +1,98 @@
+/*
+ * This file is part of the lo2s software.
+ * Linux OTF2 sampling
+ *
+ * Copyright (c) 2016,
+ *    Technische Universitaet Dresden, Germany
+ *
+ * lo2s is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * lo2s is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with lo2s.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <lo2s/config.hpp>
+#include <lo2s/cupti/events.hpp>
+#include <lo2s/log.hpp>
+#include <lo2s/perf/time/converter.hpp>
+#include <lo2s/ringbuf.hpp>
+#include <lo2s/trace/trace.hpp>
+#include <lo2s/types.hpp>
+
+#include <chrono>
+#include <cstdlib>
+#include <string>
+
+extern "C"
+{
+#include <sys/timerfd.h>
+#include <unistd.h>
+}
+
+namespace lo2s
+{
+namespace cupti
+{
+
+class Reader
+{
+public:
+    Reader(trace::Trace& trace, Process process)
+    : process_(process), trace_(trace), time_converter_(perf::time::Converter::instance()),
+      ringbuf_reader_("cupti", process.as_pid_t(), true, config().nvidia_ringbuf_size),
+      timer_fd_(timerfd_from_ns(config().userspace_read_interval)),
+      executable_name_(get_process_exe(process))
+    {
+    }
+
+    void read()
+    {
+        struct event_header* header = nullptr;
+
+        while ((header = reinterpret_cast<struct event_header*>(
+                    ringbuf_reader_.get(sizeof(struct event_header)))) != nullptr)
+        {
+            if (header->type == EventType::CUPTI_KERNEL)
+            {
+                struct event_kernel* kernel =
+                    reinterpret_cast<struct event_kernel*>(ringbuf_reader_.get(header->size));
+
+                auto& writer = trace_.cuda_writer(Thread(process_.as_thread()));
+
+                std::string kernel_name = kernel->name;
+                auto& cu_cctx = trace_.cuda_calling_context(executable_name_, kernel_name);
+
+                writer.write_calling_context_enter(time_converter_(kernel->start), cu_cctx.ref(),
+                                                   2);
+                writer.write_calling_context_leave(time_converter_(kernel->end), cu_cctx.ref());
+            }
+
+            ringbuf_reader_.pop(header->size);
+        }
+    }
+
+    int fd()
+    {
+        return timer_fd_;
+    }
+
+private:
+    Process process_;
+    trace::Trace& trace_;
+    perf::time::Converter& time_converter_;
+    RingBufReader ringbuf_reader_;
+    int timer_fd_;
+    std::string executable_name_;
+};
+} // namespace cupti
+} // namespace lo2s
--- a/include/lo2s/ringbuf.hpp
+++ b/include/lo2s/ringbuf.hpp
@@ -1,0 +1,245 @@
+/*
+ * This file is part of the lo2s software.
+ * Linux OTF2 sampling
+ *
+ * Copyright (c) 2024,
+ *    Technische Universitaet Dresden, Germany
+ *
+ * lo2s is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * lo2s is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with lo2s.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <lo2s/error.hpp>
+#include <lo2s/shared_memory.hpp>
+
+#include <atomic>
+#include <cassert>
+#include <cstddef>
+#include <cstring>
+#include <string>
+#include <system_error>
+
+extern "C"
+{
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <unistd.h>
+}
+
+namespace lo2s
+{
+
+// To resolve possible ringbuf format incompatibilities
+#define RINGBUF_VERSION 1
+
+struct ringbuf_header
+{
+    uint64_t version;
+    uint64_t size;
+    std::atomic_uint64_t head;
+    std::atomic_uint64_t tail;
+};
+
+class ShmRingbuf
+{
+public:
+    ShmRingbuf(std::string component, pid_t pid, bool create, size_t pages)
+    {
+        std::string filename = "/lo2s-" + component + "-" + std::to_string(pid);
+
+        fd_ = shm_open(filename.c_str(), create ? O_RDWR | O_CREAT | O_EXCL : O_RDWR, 0600);
+        if (fd_ == -1)
+        {
+            throw std::system_error(errno, std::system_category());
+        }
+
+        size_t pagesize = sysconf(_SC_PAGESIZE);
+        size_t size;
+
+        if (create)
+        {
+            size = pagesize * pages;
+            ftruncate(fd_, size + sysconf(_SC_PAGESIZE));
+        }
+        else
+        {
+            auto header_map = SharedMemory(fd_, sizeof(struct ringbuf_header), 0);
+            size = header_map.as<struct ringbuf_header>()->size;
+        }
+
+        // To handle events that wrap around the ringbuffer, map it twice into virtual memory
+        // back-to-back. This way events that wrap around the ringbuffer can be read and written
+        // without noticing the wraparound:
+        //
+        // in physical memory: [ent|-----|ev]
+        //
+        // in virtual memory:  [ent|-----|ev][ent----|ev]
+        //
+        // As there is no way to reserve a range of virtual memory, mmap()-ing two adjacent
+        // ring-buffer without races is tricky. We solve this problem by mmap()-ing an area twice
+        // the size of the ringbuffer and then overwriting the latter half of this mapping with
+        // another mapping of the ringbuffer using MMAP_FIXED. This way we only touch mappings we
+        // control. Also, put the ringbuffer header on a separate page to make life easier.
+
+        first_mapping_ = SharedMemory(fd_, size * 2 + pagesize, 0);
+
+        second_mapping_ = SharedMemory(fd_, size, pagesize, first_mapping_.as<std::byte>() + size);
+
+        header_ = first_mapping_.as<struct ringbuf_header>();
+        start_ = first_mapping_.as<std::byte>() + pagesize;
+
+        if (create)
+        {
+            header_->version = RINGBUF_VERSION;
+            header_->size = size;
+            header_->tail.store(0);
+            header_->head.store(0);
+        }
+        else
+        {
+            if (header_->version != RINGBUF_VERSION)
+            {
+                throw new std::runtime_error("Incompatible RingBuffer Version " +
+                                             std::to_string(header_->version) +
+                                             " detected on other side!");
+            }
+        }
+    }
+
+    uint64_t head()
+    {
+        return header_->head.load();
+    }
+
+    uint64_t tail()
+    {
+        return header_->tail.load();
+    }
+
+    void head(uint64_t new_head)
+    {
+        return header_->head.store(new_head);
+    }
+
+    void tail(uint64_t new_tail)
+    {
+        return header_->tail.store(new_tail);
+    }
+
+    uint64_t size()
+    {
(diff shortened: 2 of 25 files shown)


src/monitor/process_monitor_main.cpp
```diff
@@ -82,14 +92,22 @@ namespace monitor
...
                    });
     tmp.push_back(nullptr);
+    std::transform(env.begin(), env.end(), std::back_inserter(c_env),
+                   [](const std::string& s)
+                   {
+                       char* pc = new char[s.size() + 1];
+                       std::strcpy(pc, s.c_str());
+                       return pc;
+                   });
+    c_env.push_back(nullptr);
```
bmario, line 102: Extract these c hacking transform things into a function.

src/trace/trace.cpp
```diff
@@ -489,6 +503,15 @@ otf2::writer::local& Trace::create_metric_writer(const std::string& name)
     return archive_(location);
 }
 
+otf2::definition::calling_context& Trace::cuda_calling_context(std::string& file,
+                                                               std::string& function)
+{
+    LineInfo info = LineInfo::for_function(file.c_str(), function.c_str(), 0, "");
```
bmario, line 509: Can we get a line number from cupti?

include/lo2s/cupti/reader.hpp
```diff
@@ -0,0 +1,112 @@
...
+class Reader
+{
+public:
+    Reader(trace::Trace& trace, Process process)
+    : process_(process), trace_(trace), time_converter_(perf::time::Converter::instance()),
+      ringbuf_reader_(std::to_string(process.as_pid_t()), 512)
+    {
+        struct itimerspec tspec;
+        memset(&tspec, 0, sizeof(struct itimerspec));
+        tspec.it_value.tv_nsec = 1;
```
bmario, line 55:
```suggestion
```

include/lo2s/cupti/reader.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+        tspec.it_interval.tv_sec =
+            std::chrono::duration_cast<std::chrono::seconds>(config().userspace_read_interval)
+                .count();
+
+        tspec.it_interval.tv_nsec =
+            (config().userspace_read_interval % std::chrono::seconds(1)).count();
```
bmario, line 62:
```suggestion
        tspec.it_interval.tv_nsec =
            (config().userspace_read_interval % std::chrono::seconds(1)).count();

```

include/lo2s/cupti/reader.hpp
```diff
@@ -0,0 +1,112 @@
...
+        timer_fd_ = timerfd_create(CLOCK_MONOTONIC, TFD_NONBLOCK);
+
+        timerfd_settime(timer_fd_, TFD_TIMER_ABSTIME, &tspec, NULL);
```
bmario, line 65: This is the third reincarnation of building a tspec and using timerfd_settime. Refactor all these into a function taking a std::chrono::duration instead.

include/lo2s/cupti/reader.hpp
```diff
@@ -0,0 +1,112 @@
...
+{
+namespace cupti
+{
+
+class Reader
+{
+public:
+    Reader(trace::Trace& trace, Process process)
+    : process_(process), trace_(trace), time_converter_(perf::time::Converter::instance()),
+      ringbuf_reader_(std::to_string(process.as_pid_t()), 512)
```
bmario, line 51: just using the the pid as filename. The manpage says this:

> For portable use, a shared memory object should be identified  by  a  name  of the form /somename; that is, a null-terminated string of up to NAME_MAX (i.e., 255) characters consisting of an initial slash, followed by one or more characters, none of which are slashes.

Also, it should contain lo2s and cupti at some point.

bmario, line 51: Hard-coded size, make it a command-line option.

include/lo2s/cupti/reader.hpp
```diff
@@ -0,0 +1,112 @@
...
+        return timer_fd_;
+    }
+
+private:
+    Process process_;
+    trace::Trace& trace_;
+    perf::time::Converter& time_converter_;
+    RingBufReader ringbuf_reader_;
+    int timer_fd_;
+    std::string exe;
```
bmario, line 109:
```suggestion
    std::string executable_name_;
```

include/lo2s/cupti/reader.hpp
```diff
@@ -0,0 +1,112 @@
...
+            struct cupti_event_kernel* kernel = (struct cupti_event_kernel*)kernel_rb.get();
+
+            auto& writer = trace_.cuda_writer(Thread(process_.as_thread()));
+
+            std::string kernel_name = kernel->name;
+            auto& cu_cctx = trace_.cuda_calling_context(exe, kernel_name);
+
+            writer << otf2::event::calling_context_enter(time_converter_(kernel->start), cu_cctx,
+                                                         2);
+            writer << otf2::event::calling_context_leave(time_converter_(kernel->end), cu_cctx);
```
bmario, line 92: Use `writer.write_calling_context_enter/leave()` instead.

include/lo2s/cupti/reader.hpp
```diff
@@ -0,0 +1,112 @@
...
+    void read()
+    {
+        rb_ptr header_rb = ringbuf_reader_.get(sizeof(struct cupti_event_header));
+
+        if (!header_rb)
+        {
+            return;
+        }
+
+        struct cupti_event_header* header = (struct cupti_event_header*)header_rb.get();
```
bmario, line 78: evil c cast

include/lo2s/cupti/reader.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+        if (header->type == CuptiEventType::CUPTI_KERNEL)
+        {
+            rb_ptr kernel_rb = ringbuf_reader_.get(header->size);
+            struct cupti_event_kernel* kernel = (struct cupti_event_kernel*)kernel_rb.get();
```
bmario, line 83: evil c cast

include/lo2s/cupti/ringbuf.hpp
```diff
@@ -0,0 +1,254 @@
...
+    uint64_t version;
+    uint64_t size;
+    std::atomic_uint64_t head;
+    std::atomic_uint64_t tail;
+    std::atomic_uint64_t fill;
+};
+
+enum class CuptiEventType : uint64_t
+{
+    CUPTI_KERNEL = 0,
```
bmario, line 53: I wouldn't choose 0 as first type index but at least 1.

include/lo2s/cupti/ringbuf.hpp
```diff
@@ -0,0 +1,254 @@
...
+};
+
+struct cupti_event_header
+{
+    CuptiEventType type;
+    uint64_t size;
+};
```
bmario, line 60: I think, besides `CUPTI_KERNEL`, there's no need to smurf this with cupti here.

include/lo2s/cupti/ringbuf.hpp
```diff
@@ -0,0 +1,254 @@
...
+        ringbuf_->fill.fetch_add(reserved_size_);
+
+        reserved_ = nullptr;
+        reserved_size_ = 0;
+
+        return true;
+    }
+
+private:
+    std::string filename_;
```
bmario, line 173: unused

include/lo2s/cupti/ringbuf.hpp
```diff
@@ -0,0 +1,254 @@
...
+public:
+    RingBufWriter(std::string filename, size_t size) : filename_(filename)
+    {
+        fd_ = shm_open(filename.c_str(), O_RDWR, 0600);
+        if (fd_ == -1)
+        {
+            throw std::system_error(errno, std::system_category());
+        }
+
+        ringbuf_ = (struct ringbuf_header*)mmap(NULL, sizeof(struct ringbuf_header) + size,
```
bmario, line 111: evil c ast

src/cupti/lib.cpp
bmario, line 1: This file need comments. I have no idea what is happening here.

bmario (changes requested):

include/lo2s/cupti/ringbuf.hpp
```diff
@@ -0,0 +1,254 @@
...
+public:
+    RingBufReader(std::string filename, size_t size)
+    {
+        fd_ = shm_open(filename.c_str(), O_CREAT | O_EXCL | O_RDWR, 0600);
+        if (fd_ == -1)
+        {
+            throw std::system_error(errno, std::system_category());
+        }
+
+        ringbuf_ = (struct ringbuf_header*)mmap(NULL, sizeof(struct ringbuf_header) + size,
```
bmario, line 192: There is `EventReader::init_mmap()`, which almost does the same thing, but the right way.

bmario (changes requested):
After disccussing the ringbuffer a bit with @tilsche, we think it would be better to not let individual events split up on the wrap up, thus removing the need for all the mallocs, tracking thereof and simplifying the code at the same time. Also, `fill` shouldn't be necessary.

include/lo2s/cupti/ringbuf.hpp
cvonelm: We either need `fill` or to leave one byte empty between head and tail, because `head == tail` could mean both full and empty. And I personally prefer fill over leaving one byte free because the latter feels like it would invite off-by-one errors

src/cupti/lib.cpp
```diff
@@ -0,0 +1,199 @@
...
+    }
+    catch (std::exception const& e)
+    {
+        std::cerr << "Could not read lo2s ringbuffer size!";
+        exit(EXIT_FAILURE);
+    }
+
+    atexit(&atExitHandler);
+
+    cuptiSubscribe(&subscriber, (CUpti_CallbackFunc)callbackHandler, NULL);
```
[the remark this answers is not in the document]
cvonelm: the call creates  the subscriber, so  we cant pass it as userData to the same cupti call, or am I missing something

src/monitor/process_monitor.cpp
```diff
@@ -38,10 +38,11 @@ void ProcessMonitor::insert_process(Process parent, Process process, std::string
                                     bool spawn)
 {
     trace_.add_process(parent, process, proc_name);
-    insert_thread(process, process.as_thread(), proc_name, spawn);
+    insert_thread(process, process.as_thread(), proc_name, spawn, true);
 }
 
-void ProcessMonitor::insert_thread(Process process, Thread thread, std::string name, bool spawn)
+void ProcessMonitor::insert_thread(Process process, Thread thread, std::string name, bool spawn,
+                                   bool is_process)
```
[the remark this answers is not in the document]
cvonelm: The ScopeMonitor is created in insert_thread. And CUDA sampling should only be set up for processes not threads

src/trace/trace.cpp
cvonelm: nope

include/lo2s/cupti/ringbuf.hpp
tilsche: I'd rather "waste" one byte in the buffer than have one atomic 8-byte header thing to take care of.
You take care of off-by-one errors with a proper function to count the free space.

See also https://www.kernel.org/doc/Documentation/circular-buffers.txt

P.S. I do quite frankly wonder why one would not just keep incrementing head/tail indefinitely and only `%` them upon access rather than wrap-around. That would seem to be more elegant to me - and you could use the precious last byte! But I cannot oversee the consequences and would suggest to rather go with the proven approach.

bmario (changes requested):
I have these comments still pending. Don't know if they still apply.

include/lo2s/ringbuf.hpp
```diff
@@ -0,0 +1,271 @@
...
+        header_ = first_mapping_.as<struct ringbuf_header*>();
+        start_ = first_mapping_.as<std::byte*>() + pagesize;
+
+        if (create)
+        {
+            header_->version = RINGBUF_VERSION;
+            header_->size = ringbuf_size;
+            header_->tail.store(0);
+            header_->head.store(0);
+        }
```
bmario, line 163: ringbuf version is never checked.

include/lo2s/ringbuf.hpp
```diff
@@ -0,0 +1,271 @@
...
+    uint64_t head()
+    {
+        return header_->head.load();
+    }
+
+    uint64_t tail()
+    {
+        return header_->tail.load();
+    }
+    void head(uint64_t new_head)
```
bmario, line 174:
```suggestion

    void head(uint64_t new_head)
```

include/lo2s/ringbuf.hpp
```diff
@@ -0,0 +1,271 @@
...
+    {
+        return header_->head.store(new_head);
+    }
+
+    void tail(uint64_t new_tail)
+    {
+        return header_->tail.store(new_tail);
+    }
+
+    uint64_t ringbuf_size()
```
bmario, line 184:
```suggestion
    uint64_t size()
```

include/lo2s/ringbuf.hpp
```diff
@@ -0,0 +1,271 @@
...
+
+    ~Mmap()
+    {
+        munmap(addr_, size_);
+    }
+
+private:
+    void* addr_ = nullptr;
+    size_t size_;
+};
```
bmario, line 106: Extract this as a separate PR and replace `EventReader::init_mmap()` with it?

bmario: At the very least, extract into separate file

bmario (approved):

bmario (changes requested):
I blame vscode

include/lo2s/ringbuf.hpp
cvonelm: https://github.com/tud-zih-energy/lo2s/pull/341

include/lo2s/ringbuf.hpp
```diff
@@ -0,0 +1,230 @@
...
+        if (!can_be_loaded(size))
+        {
+            return nullptr;
+        }
+        return start_ + tail();
+    }
+
+    void pop(size_t ev_size)
+    {
+        // Calling pop() without trying to get() data from the ringbuffer first is a error
```
bmario, line 213:
```suggestion
        // Calling pop() without trying to get() data from the ringbuffer first is an error
```

bmario (changes requested):

src/trace/trace.cpp
```diff
@@ -411,6 +410,21 @@ otf2::writer::local& Trace::sample_writer(const ExecutionScope& writer_scope)
     return archive_(location(writer_scope));
 }
 
+otf2::writer::local& Trace::cuda_writer(const Thread& thread)
+{
+    MeasurementScope scope = MeasurementScope::cuda(thread.as_scope());
+
+    const auto& cuda_location_group = registry_.emplace<otf2::definition::location_group>(
+        ByMeasurementScope(scope), intern(scope.name()), otf2::common::location_group_type::process,
```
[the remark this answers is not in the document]
cvonelm: I have changed those to accelerator_(stream) and while it does not seem to change anything in vampir, it does seem to be more correct.

include/lo2s/ringbuf.hpp
```diff
@@ -0,0 +1,230 @@
...
+    RingBufWriter(std::string component, pid_t pid, bool create, size_t pages = 0)
+    : ShmRingbuf(component, pid, create, pages)
+    {
+    }
+
+    std::byte* reserve(size_t ev_size)
+    {
+        // No other reservation can be active!
+        assert(reserved_size_ == 0);
+
```
bmario, line 168: if ev_size == 0 return nullptr

include/lo2s/ringbuf.hpp
```diff
@@ -0,0 +1,230 @@
...
+{
+public:
+    RingBufReader(std::string component, pid_t pid, bool create, size_t pages = 0)
+    : ShmRingbuf(component, pid, create, pages)
+    {
+    }
+
+    std::byte* get(size_t size)
+    {
+        if (!can_be_loaded(size))
```
bmario, line 204: if size == 0 return nullptr

include/lo2s/ringbuf.hpp
```diff
@@ -0,0 +1,230 @@
...
+        {
+            return nullptr;
+        }
+        return start_ + tail();
+    }
+
+    void pop(size_t ev_size)
+    {
+        // Calling pop() without trying to get() data from the ringbuffer first is an error
```
bmario, line 213: if size ==0 return (or assert size > 0)

src/monitor/process_monitor_main.cpp
```diff
@@ -117,6 +119,19 @@ static void drop_privileges()
     assert(getgid() != 0);
 }
 
+std::vector<char*> to_vector_of_c_str(std::vector<std::string> vec)
```
bmario, line 122:
```suggestion
std::vector<char*> to_vector_of_c_str(const std::vector<std::string>& vec)
```

bmario (changes requested):

bmario (approved):

PR #325, opened 2024-03-13
fix: CPU metric has value wrong value

### LINKED ISSUE #324: CPU metric shows wrong value
Not scheduled metrics are supposed to have `-1` as value for CPU, but it is 2³² - 1 instead.

Vampir screenshot:
grafik

bmario:
For not scheduled threads/processes, the CPU should show -1, but due to implicit casting to unsigned, it used UINT_MAX instead.

Fixes #324


tilsche (approved):
Change looks, but the similarity between the two methods is striking...

PR #331, opened 2024-04-10
Initialize arrays in lo2s::Pipe

cvonelm:
More modern compilers complain about fds_ and fd_open_ being uninitialized, so initialize them


tilsche (approved):
lgtm

Issue #301, opened 2023-09-07
setuid() support for process mode


cvonelm:
Currently, the program under measurement runs with the same privileges as lo2s itself. This is problematic when the measurement feature you want to use (upcoming POSIX I/O support for example) is only available to root and you can't or don't want to run the program under measurement as root.

teto519f:
I've looked into this a bit and as far as I can see, there isn't a perfect way to implement this.

The actual dropping of privileges is more or less simple, see https://www.oreilly.com/library/view/secure-programming-cookbook/0596003943/ch01s03.html

But simply adding a flag, which has to be added when calling on lo2s with higher privileges, isn't pretty, since it could be forgotten or deliberately omitted.

Everything would work just fine if we assume the user gained their privileges through sudo.
Sudo changes effective, real and saved uid to 0, which we could check for and reset them to the actual users uid before then calling on exec(). The actual user id is saved in the environment var SUDO_UID and the actual group id in SUDO_GID (see https://www.geeksforgeeks.org/real-effective-and-saved-userid-in-linux/ and https://man7.org/linux/man-pages/man8/sudo.8.html).

But what if sudo wasn't used? I haven't found a way to access the callers id in that case, since all variants of getuid() or getgid() would access the already changed ids (see https://man7.org/linux/man-pages/man2/getuid.2.html or corresponding man pages).

s9105947:
thx for the writeup.

We have checked back with some more colleagues, which recommended strongly against using the setuid flag on the lo2s binary itelf, and recommended using sudo.
The rationale is that setuid opens vectors through e.g. environment variables. sudo cleans up the environment and prevents this particular issue.
I haven't found a concise writeup on suid binaries vs sudo from brief googling, so maybe this is overly cautious.
Note that the "proper way" to do suid is to have a component that does **only** do the stuff that requires privileges, and another component that does all other.

cvonelm:
I think I have worded this ambiguously.

What I want is root -> start lo2s -> fork -> setuid() to normal user -> start program under measurement as normal user. (Kind of how webservers acquire low port numbers and then drop to a less privileged user for the actual serving)

I agree that using the setuid-*bit* is a security disaster waiting to happen and something that we will not really be able to convince people to just deploy on their systems.

Something that i've tried some time ago is 

```bash
$ sudo lo2s -- sudo -u normal_user ./application
``` 

[...]

PR #333, opened 2024-05-24
Drop root privileges before forking

teto519f:
Implemented a way for lo2s to detect a call with sudo and then drop its privileges before launching any process. Different methods of gaining root aren't taken into account (yet).

This refers to #301 

To Do:

- [x] test functionality in real use case
- [x] ensure compatibility with lo2s -a
- [x] look into doas support (then update manpage)

### DIFF
--- a/man/lo2s.1.pod
+++ b/man/lo2s.1.pod
@@ -154,6 +154,14 @@ determine the trace path instead, including variable substitution.
 Attach to a running process with process ID I<PID> instead of launching
 I<COMMAND>.
 
+=item B<-u>, B<--as-pre-sudo-user>
+
+Launch I<COMMAND> as the user that called on sudo. Requires a lo2s call with sudo.
+
+=item B<-U>, B<--as-user> I<USERNAME>
+
+Launch I<COMMAND> as I<USERNAME>. The caller needs to have CAP_SETUID. Will take priority over B<-u> if specified together.
+
 =item B<-m>, B<--mmap-pages> I<N> (default: C<16>)
 
 Allocate I<N> pages for each internal buffer shared between B<lo2s> and the
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -185,6 +185,14 @@ void parse_program_options(int argc, const char** argv)
         .metavar("PID")
         .optional();
 
+    general_options.toggle("as-pre-sudo-user", "Drop root privileges before launching COMMAND.")
+        .short_name("u");
+
+    general_options.option("as-user", "User to drop privileges to before launching COMMAND.")
+        .short_name("U")
+        .metavar("USERNAME")
+        .optional();
+
     general_options.option("mmap-pages", "Number of pages to be used by internal buffers.")
         .short_name("m")
         .default_value("16")
@@ -362,6 +370,7 @@ void parse_program_options(int argc, const char** argv)
     config.mmap_pages = arguments.as<std::size_t>("mmap-pages");
     config.process =
         arguments.provided("pid") ? Process(arguments.as<pid_t>("pid")) : Process::invalid();
+    config.drop_root = arguments.given("as-pre-sudo-user");
     config.sampling_event = arguments.get("event");
     config.sampling_period = arguments.as<std::uint64_t>("count");
     config.enable_cct = arguments.given("call-graph");
@@ -419,6 +428,18 @@ void parse_program_options(int argc, const char** argv)
                 break;
             }
         }
+    }
+
+    if (arguments.given("as-pre-sudo-user") && std::getenv("SUDO_UID") == NULL)
+    {
+        Log::error() << "-u was specified but no sudo was detected.";
+        std::exit(EXIT_FAILURE);
+    }
+
+    if (arguments.provided("as-user"))
+    {
+        config.drop_root = true;
+        config.user = arguments.get("as-user");
     }
 
     // list arguments to arguments and exit
--- a/src/monitor/process_monitor_main.cpp
+++ b/src/monitor/process_monitor_main.cpp
@@ -41,6 +41,8 @@
 
 extern "C"
 {
+#include <grp.h>
+#include <pwd.h>
 #include <signal.h>
 #include <unistd.h>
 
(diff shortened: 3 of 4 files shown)


teto519f:
With current changes you cannot call on any process, that needs sudo through lo2s. 
For example `lo2s -- sudo apt update` does not work, since lo2s can't call on sudo without root privileges.
`sudo lo2s -- sudo apt update` should be something you're able to do (imo), but this doesn't work when dropping privileges.

TL;DR I would like to add an option, that disables the sudo/doas check. Is this valid?

cvonelm:
> TL;DR I would like to add an option, that disables the sudo/doas check. Is this valid?

I would do it the other way around, enabling this getuid functionality only if some option is given. This feels like something that might lead to weird-to-detect errors if lo2s would just do privilege dropping automatically

Ideally, this option would also accept some username, so that you could specify that you want to switch to that specific user.

You can get the UID/GID of some username by using getpwnam (https://linux.die.net/man/3/getpwnam)

man/lo2s.1.pod
```diff
@@ -154,6 +154,14 @@ determine the trace path instead, including variable substitution.
 Attach to a running process with process ID I<PID> instead of launching
 I<COMMAND>.
 
+=item B<-u>, B<--as-current-user>
+
+Launch I<COMMAND> as the current user. Requires lo2s call with sudo.
+
+=item B<-U>, B<--as-user> I<USERNAME>
+
+Launch I<COMMAND> as the specified user. Implies B<-u> but does not require sudo, but the current user has to have setuid() privileges. Will take priority over B<-u> if specified together.
```
cvonelm, line 163: user I<USERNAME> instead of "the specified user" . I think the part about "implies" in its current implementation confuses more than it helps

src/config.cpp
```diff
@@ -185,6 +185,15 @@ void parse_program_options(int argc, const char** argv)
         .metavar("PID")
         .optional();
 
+    general_options
+        .toggle("as-current-user", "Drop root privileges to current user before launching COMMAND.")
+        .short_name("u");
+
+    general_options.option("as-user", "Specify a user to drop privileges to.")
```
cvonelm, line 192: change wording in accordance with man page comment above.

src/config.cpp
```diff
@@ -421,6 +431,18 @@ void parse_program_options(int argc, const char** argv)
         }
     }
 
+    if (arguments.given("as-current-user") && std::getenv("SUDO_UID") == NULL)
+    {
+        Log::error() << "-u was specified but no sudo was used.";
```
cvonelm, line 436: used -> detected

man/lo2s.1.pod
```diff
@@ -154,13 +154,13 @@ determine the trace path instead, including variable substitution.
...
-=item B<-u>, B<--as-current-user>
+=item B<-u>, B<--as-pre-sudo-user>
 
-Launch I<COMMAND> as the current user. Requires lo2s call with sudo.
+Launch I<COMMAND> as the user that called on sudo. Requires a lo2s call with sudo.
 
 =item B<-U>, B<--as-user> I<USERNAME>
 
-Launch I<COMMAND> as the specified user. Implies B<-u> but does not require sudo, but the current user has to have setuid() privileges. Will take priority over B<-u> if specified together.
+Launch I<COMMAND> as I<USERNAME>. I<USERNAME> has to have setuid() privileges. Will take priority over B<-u> if specified together.
```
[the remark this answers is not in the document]
teto519f: Do you think "the caller" or "the calling user" is a good description for this user? Just wanted to ask before I commit again.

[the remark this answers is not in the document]
cvonelm: "the caller needs to have CAP_SETUID" is the most correct way to write it me thinks

cvonelm (approved):

cvonelm:
I have tested this now with our main use-case (which is running the IOR benchmark with mpirun, while lo2s runs as root) which seems to run fine, so we can check of testing

src/config.cpp
```diff
@@ -185,6 +185,14 @@ void parse_program_options(int argc, const char** argv)
         .metavar("PID")
         .optional();
 
+    general_options.toggle("as-pre-sudo-user", "Drop root privileges before launching COMMAND.")
```
bmario, line 188: The internal config option is called `drop_root`, the description is "Drop root privileges ...", but the thing is called `--as-pre-sudo-user`. I think that is confusing. Don't differ between "internal" and "external" naming of features. I think `--drop-root(-privileges)` would have been a better name.

src/config.cpp
```diff
@@ -185,6 +185,14 @@ void parse_program_options(int argc, const char** argv)
...
+        .short_name("u");
+
+    general_options.option("as-user", "User to drop privileges to before launching COMMAND.")
```
bmario, line 191: I think this is easier to understand:
"Launch the COMMAND as the user USERNAME"

src/config.cpp
```diff
@@ -421,6 +430,18 @@ void parse_program_options(int argc, const char** argv)
         }
     }
 
+    if (arguments.given("as-pre-sudo-user") && std::getenv("SUDO_UID") == NULL)
```
bmario, line 433: This is not the behavior I would expect.
While `lo2s -a -A` works fine, `lo2s -u -U $USER` complains about `-u was specified but no sudo was detected.`

Or in other words, `-U` should imply `-u`.

bmario: Also, in C++, we don't use `NULL`. Use `nullptr` instead.

src/monitor/process_monitor_main.cpp
```diff
@@ -53,6 +55,72 @@ namespace lo2s
...
+            Log::error() << "No user found with username '" + config().user + ".";
+            throw_errno();
+        }
+
+        original_uid = orig_user->pw_uid;
+        original_gid = orig_user->pw_gid;
+    }
+    else
+    {
+        // sudo is garanteed in this case, see check in config.hpp
```
bmario, line 78: Don't only put your assumptions into comments. Put them in `assert()`s.

src/monitor/process_monitor_main.cpp
```diff
@@ -53,6 +55,72 @@ namespace lo2s
...
+        try
+        {
+            original_uid = std::stoi(std::getenv("SUDO_UID"));
+            original_gid = std::stoi(std::getenv("SUDO_GID"));
+        }
+        catch (std::invalid_argument const& e)
+        {
+            Log::error() << "Cannot parse SUDO_UID/SUDO_GID into int.";
```
bmario, line 86: This will be a user-facing message. Users don't know what an `int` is or that `SUDO_UID/SUDO_GID` are environment variables.

`Cannot parse the environment variables SUDO_UID and/or SUDO_GID.`

src/monitor/process_monitor_main.cpp
```diff
@@ -53,6 +55,72 @@ namespace lo2s
...
+            throw_errno();
```
bmario, line 87: `std::stoi` raises an exception, but does not set `errno`. `throw_errno()` would need that.

Replace with `throw;`, which rethrows the current exception.

src/monitor/process_monitor_main.cpp
```diff
@@ -53,6 +55,72 @@ namespace lo2s
...
+        }
+        catch (std::out_of_range const& e)
```
bmario, line 89: I don't think the distinction between both exceptions types adds any value here. I'd rather have one "parsing failed" message.

src/monitor/process_monitor_main.cpp
```diff
@@ -53,6 +55,72 @@ namespace lo2s
...
+        {
+            Log::error() << "SUDO_UID/SUDO_GID out of range.";
+            throw_errno();
```
bmario, line 92: `throw_errno()` doesn't work here, but remove it completely.

src/monitor/process_monitor_main.cpp
teto519f: Would you remove the second catch statement and catch general exceptions or duplicate the error handling for both types? 
I'd go for the second option, but redundant code is wack and a function would be overkill.

bmario: I'd go with one handler `catch(const std::exception& e)`. As users typically do not set these environment variables themselves, I don't think any other message than "failed to parse ..."  is necessary, but rather confusing.

Issue #311, opened 2023-12-11
Advertise --userspace-metric-event more


cvonelm:
userspace-metric-event uses perf with all the features that can produce problems with special kinds of events disabled, such as event grouping and using ring-buffers for event read-out. As this, it should be considered the fallback option to -E.

However, it is nowhere advertised in lo2s that you might want to try --userspace-metric-event if -E doesnt work for you.

bmario:
In addition, add it to the manpage

cvonelm:
```
   Metric options
       -E, --metric-event EVENT
           Record metrics for this perf event.  May be specified multiple
           times to record metrics for more than one event.  Try
           --userspace-metric-event if EVENT is not openable.

       --userspace-metric-event EVENT
           This is a more compatible but slower version of -E.
```

This is already mentioned in the manpage

PR #336, opened 2024-06-06
Point to --userspace-metric-event

teto519f:
Since the "failed to add counter ..." error seems to be the one occurring if -E does not work, I decided to add a hint to --userspace-metric-event there.

This solves #311


bmario:
Please also update the man page to:
- include `--userspace-metric-event`
- let `-E` and `--userspace-metric-event` suggest each other

bmario (approved):

PR #337, opened 2024-06-21
fix(sample): Drop cpuid_metric event from sample writing.

### LINKED ISSUE #327: Consider dropping cpuid metric events from samples
We currently write the cpuid metric events both for scheduling events, but als at the samples themselves (see `perf::sample::Writer::cpuid_metric_event_`). The latter should usually be redundant, since it should always be encased in respective scheduling events. Are there still cases where we need to write it alongside calling context _samples_?

cvonelm:
Since 8f3ac68, we do not support systems that do not have perf_attr.context_switch anymore. This means that since then, samples will always be enclosed in context switch events, removing the need for writing the cpuid metric from samples.

Also remove src/perf/syscall/sample/writer.cpp. Apparently this is completely dead code

This fixes #327


src/perf/syscall/sample/writer.cpp
bmario, line 1: Why can we remove this file?

cvonelm: According to `git grep` this just seems to have sneaked into the repo at some point. It was never used and just looks like an accidental copy of src/perf/sample/writer.cpp

bmario (approved):

PR #339, opened 2024-07-18
Fix NEC sampling if threads are involved

cvonelm:
ve_check_pid() only works for processes, not threads, so lo2s complained that it could not find the real NEC device id for those threads.

Rewrite the code to use ve_sysfs_path_info(). This gives us the correct sysfs path for a NEC device id, so we dont have to find it using ve_check_pid().

### DIFF
--- a/src/monitor/nec_monitor_main.cpp
+++ b/src/monitor/nec_monitor_main.cpp
@@ -19,31 +19,35 @@
  * along with lo2s.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <filesystem>
+
+#include <lo2s/log.hpp>
 #include <lo2s/monitor/nec_monitor_main.hpp>
+
+extern "C"
+{
+    extern int ve_sysfs_path_info(int nodeid, char* ve_sysfs_path);
+}
 
 namespace lo2s
 {
 namespace nec
 {
 
-std::optional<NecDevice> NecMonitorMain::get_device_of(Thread thread)
-{
-    // /sys/class/ve/ve0 is not device 0, because that would make too much sense
-    // So look the device id up here
-
-    for (int i = 0; i < nodeinfo_.total_node_count; i++)
-    {
-        if (!ve_check_pid(nodeinfo_.nodeid[i], thread.as_pid_t()))
-        {
-            return NecDevice(nodeinfo_.nodeid[i]);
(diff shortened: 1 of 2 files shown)


src/monitor/nec_monitor_main.cpp
```diff
@@ -19,31 +19,30 @@
...
-    return std::optional<NecDevice>();
-}
-
 std::vector<Thread> NecMonitorMain::get_tasks_of(NecDevice device)
 {
-    std::ifstream task_stream(fmt::format("/sys/class/ve/ve{}/task_id_all", device.as_int()));
+    char sysfs_path_str[PATH_MAX];
+    memset(sysfs_path_str, 0, PATH_MAX);
+
+    ve_sysfs_path_info(device.as_int(), sysfs_path_str);
```
bmario, line 42: return ignored.

Also, I suspect you could do the old `auto size = ve_sysfs_path_info(device.as_int(), sysfs_path_str);` to get the size of the string?

src/monitor/nec_monitor_main.cpp
```diff
@@ -19,31 +19,30 @@
...
+    std::filesystem::path sysfs_path(sysfs_path_str);
+    sysfs_path /= "task_id_all";
```
bmario, line 44: Technically speaking, if sysfs_path_str is allowed to be PATH_MAX, appending `/task_id_all` here produces a path potentially longer when PATH_MAX.

bmario (changes requested):

src/monitor/nec_monitor_main.cpp
cvonelm: This just returns 0 on success, -1 on failure.

This is all the information that exists on that function in the world:
```
# declared in internal include file, but useful little helper function
cdef extern int ve_sysfs_path_info(int nodeid, char *ve_sysfs_path)
```

I just throw_errno() on -1, as this seems to set errno.

src/monitor/nec_monitor_main.cpp
cvonelm: I think that is something we have to live with. Hopefully nobody is putting enough NEC VE's into a single computer to outrun PATH_MAX.

bmario (approved):

PR #340, opened 2024-07-23
Turn EventDescription into class

teto519f:
This addresses issue #224 

This PR creates a PerfEvent and a PerfEventInstance class, which can handle every type of event in lo2s. It aims to remove redundant code for event creation and to simplify their handling.


include/lo2s/perf/event_provider.hpp
```diff
@@ -82,14 +45,14 @@ class EventProvider
         return instance_mutable();
     }
 
-    static EventDescription get_event_by_name(const std::string& name);
+    static SysfsEvent get_event_by_name(const std::string& name);
```
cvonelm, line 48: SysfsEvent -> PerfEvent for everything besides get_pmu_events (which uses sysfs and thus can only return SysfsEvents)

include/lo2s/perf/io_reader.hpp
```diff
@@ -92,15 +93,13 @@ class IoReader : public PullReader
 public:
     IoReader(IoReaderIdentity identity) : identity_(identity)
     {
-        struct perf_event_attr attr = common_perf_event_attrs();
-        attr.type = PERF_TYPE_TRACEPOINT;
+        perf::TracepointEvent event(0, identity.tracepoint.id());
```
cvonelm, line 96: We are already in  the perf namespace, so the perf:: shouldnt be necesssary

src/platform.cpp
```diff
@@ -217,7 +217,7 @@ Processor detect_processor(void)
     return Processor::UNKNOWN;
 }
 
-std::vector<perf::EventDescription> get_mem_events()
+std::vector<perf::SysfsEvent> get_mem_events()
```
cvonelm, line 220: Sysfs -> PerfEvent

src/perf/reader.cpp
```diff
@@ -0,0 +1,344 @@
...
+    attr_.bp_addr = addr;
+#else
+    attr_.type = PERF_TYPE_HARDWARE;
+    attr_.config = PERF_COUNT_HW_INSTRUCTIONS;
+    attr_.sample_period = 100000000;
+    attr_.task = 1;
+#endif
+}
+
+TracepointEvent::TracepointEvent(bool enable_on_exec, int event_id) : PerfEvent(enable_on_exec)
```
cvonelm, line 253: I  _feel_ there should be a TracepointEvent constructor that just takes the name of the tracepoint and does the tracepoint parsing stuff internally.

include/lo2s/perf/counter/counter_collection.hpp
```diff
@@ -33,18 +33,18 @@ namespace counter
 {
 struct CounterCollection
 {
-    EventDescription leader;
-    std::vector<EventDescription> counters;
+    SysfsEvent leader;
```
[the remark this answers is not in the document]
teto519f: I've grouped them into SysfsEvent, because they all used the EventDescription struct before and have the same sample_type.
I'd like to keep them like this (maybe with a different name) or, if that isn't an option, create a new class for group / userspace counter- and leader-events. But putting this into the base class PerfEvent would give all other event classes unnecessary functions that they don't use.

src/perf/reader.cpp
teto519f: Should TracepointEvent then replace the whole EventFormat class?

cvonelm: ideally, yes

include/lo2s/perf/counter/counter_collection.hpp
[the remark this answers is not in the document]
cvonelm: From my pov it looked like it was using the sysfsevent code for everything, even things not necessarily coming from sysfs. If it is not strictly sysfs but also not the perfevent base class, then introduce another class.

include/lo2s/perf/reader.hpp
```diff
@@ -0,0 +1,368 @@
...
+ * @note define the type by either calling as_counter() or as_group_leader() after creating an
+ * object
+ */
+class PredefEvent : public PerfEvent
+{
+public:
+    PredefEvent();
+    PredefEvent(bool enable_on_exec, const std::string& ev_name);
+    PredefEvent(std::string name, perf_type_id type, std::uint64_t config,
+                std::uint64_t config1 = 0, std::set<Cpu> cpus = std::set<Cpu>());
```
cvonelm, line 135: Check where cpus is coming from and ideally integrate that code directly into the Constructor.

src/perf/reader.cpp
```diff
@@ -0,0 +1,463 @@
...
+    else
+    {
+        if (sys_ev.get_fd() == -1)
+            availability_ = Availability::PROCESS_MODE;
+        else
+            availability_ = Availability::UNIVERSAL;
+    }
+}
+
+bool SysfsEvent::degrade_percision()
```
cvonelm, line 183: should be of PerfEvent, not SysfsEvent.

include/lo2s/perf/reader.hpp
```diff
@@ -0,0 +1,368 @@
...
+
+/**
+ * Contains an event parsed from sysfs
+ * @note call on as_sample() after creation to get a valid
+ * event, otherwise the availability will be set to UNAVAILABLE
+ */
+class SysfsEvent : public PredefEvent
+{
+public:
+    using PredefEvent::PredefEvent;
```
cvonelm, line 192: this is not necessary, or?

include/lo2s/perf/reader.hpp
```diff
@@ -0,0 +1,368 @@
...
+    std::set<Cpu> cpus_;
+    Availability availability_ = Availability::UNAVAILABLE;
+};
+
+/**
+ * Contains an event parsed from sysfs
+ * @note call on as_sample() after creation to get a valid
+ * event, otherwise the availability will be set to UNAVAILABLE
+ */
+class SysfsEvent : public PredefEvent
```
cvonelm, line 189: Move everything in sysfs_read_event (src/perf/event_provider) into the constructor of SysfsEvent

include/lo2s/perf/reader.hpp
```diff
@@ -0,0 +1,368 @@
...
+    double scale_ = 1;
+    std::string unit_ = "#";
+};
+
+/**
+ * Contains predefined events, like the ones listed by --list-events
+ * @note define the type by either calling as_counter() or as_group_leader() after creating an
+ * object
+ */
+class PredefEvent : public PerfEvent
```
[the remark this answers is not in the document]
teto519f: Since a PredefEvent has as_counter() and as_group_leader() this would also be moved into PerfEvent and it feels weird that you could then make a TimeEvent and call as_counter() on it, for example.

src/monitor/tracepoint_monitor.cpp
```diff
@@ -39,9 +39,8 @@ TracepointMonitor::TracepointMonitor(trace::Trace& trace, Cpu cpu)
     for (const auto& event_name : config().tracepoint_events)
```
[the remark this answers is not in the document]
teto519f: This does not exactly work, since calling the initializer from config means that not all config values are set (possibly) and I can't create a complete TracepointEvent. (The error that will occur is "No value set")

I could however create invalid TracepointEvents in the initializer that can at least be iterated over and replaced with a valid one when needed.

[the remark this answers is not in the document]
cvonelm: The issue is values taken from config()?

I think that it is not valid  to build the *Event classes in a way that they can only be correctly be constructed after we have a config(). This leads to many cases where the order in which things are constructed ends up with the *Event types blowing up in our faces, because some class containing a PerfEvent somewhere get constructed before the config is.

Please rewrite the code *Event code  so it does not read it directly from the config but has setters like set_foo, set_bar ...

include/lo2s/perf/reader.hpp
```diff
@@ -0,0 +1,368 @@
...
+    {
+        attr_.sample_type |= PERF_SAMPLE_IDENTIFIER;
+    }
+
+    const auto& fields() const
+    {
+        return fields_;
+    }
+
+    const auto& common_fields() const
```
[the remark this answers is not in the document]
teto519f: The common_fields_ member also seems to not be used anywhere, delete that too?
Not sure, because that would mean to throw all fields that start with "common_" away.

include/lo2s/perf/syscall/reader.hpp
```diff
@@ -68,110 +69,50 @@ class Reader : public EventReader<T>
...
+            enter_ev_.set_syscall_filter();
+            exit_ev_.set_syscall_filter();
         }
 
-        ioctl(fd_, PERF_EVENT_IOC_ID, &sys_enter_id);
-        ioctl(other_fd_, PERF_EVENT_IOC_ID, &sys_exit_id);
+        enter_ev_.enable();
+        exit_ev_.enable();
+
+        ioctl(enter_ev_.get_fd(), PERF_EVENT_IOC_ID, &sys_enter_id);
```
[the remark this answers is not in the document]
teto519f: What part? Or do you mean the whole thing?

include/lo2s/perf/reader.hpp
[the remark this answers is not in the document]
cvonelm: A quick test shows that we are not recording the fields starting with common_ anyways currently, so we  can remove it.

include/lo2s/perf/syscall/reader.hpp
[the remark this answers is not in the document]
cvonelm: The ioctl should be replaced by a get_id() method, similiar to the other ioctl wrappers in PerfEvent.

include/lo2s/perf/tracepoint/writer.hpp
```diff
@@ -57,7 +57,7 @@ class Writer : public Reader<Writer>
     bool handle(const Reader::RecordSampleType* sample);
 
 private:
-    EventFormat event_;
+    std::string name_;
```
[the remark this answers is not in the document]
teto519f: Would like to remove name_ completely, since the TracepointEvent from tracepoint::Reader is used here.

include/lo2s/perf/reader.hpp
```diff
@@ -0,0 +1,368 @@
...
+        return fd_ >= 0;
+    };
+
+    template <class T>
+    T read()
+    {
+        T val;
+        if (::read(fd_, &val, sizeof(val)) != sizeof(T))
+        {
+            throw std::system_error(errno, std::system_category());
```
bmario, line 351:
```suggestion
            throw_errno();
```

include/lo2s/perf/sample/reader.hpp
```diff
@@ -84,120 +85,61 @@ class Reader : public EventReader<T>
...
-
-                continue;
+                if (scope.is_cpu())
+                {
+                    ev_instance_ = event.open(scope.as_cpu());
+                }
+                else
+                {
+                    ev_instance_ = event.open(scope.as_thread());
+                }
```
bmario, line 103: There should be an overload of open for scope, which is exactly this code.

include/lo2s/perf/sample/reader.hpp
```diff
@@ -220,24 +162,13 @@ class Reader : public EventReader<T>
...
-            fd_ = -1;
-        }
     }
 
 protected:
     bool has_cct_;
 
 private:
-    int fd_ = -1;
+    PerfEventInstance ev_instance_;
```
bmario, line 171: Assuming you renamed the class to `EventGuard`, this member variable could simply be called `event_`.

src/perf/counter/userspace/reader.cpp
```diff
@@ -65,19 +66,57 @@ Reader<T>::Reader(ExecutionScope scope)
...
+                perf_warn_paranoid();
+
+                if (scope.is_cpu())
+                {
+                    counter = event.open(scope.as_cpu());
+                }
+                else
+                {
+                    counter = event.open(scope.as_thread());
+                }
```
bmario, line 98: use an overload of Event::open instead.

src/perf/counter/userspace/reader.cpp
```diff
@@ -65,19 +66,57 @@ Reader<T>::Reader(ExecutionScope scope)
...
+        try
+        {
+            if (scope.is_cpu())
+            {
+                counter = event.open(scope.as_cpu());
+            }
+            else
+            {
+                counter = event.open(scope.as_thread());
+            }
```
bmario, line 80: Overload

include/lo2s/perf/tracepoint/writer.hpp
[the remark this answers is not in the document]
cvonelm: That is fine too

src/perf/counter/group/reader.cpp
```diff
@@ -53,79 +53,98 @@ Reader<T>::Reader(ExecutionScope scope, bool enable_on_exec)
...
+        try
+        {
+            if (scope.is_cpu())
+            {
+                counter_leader_ = counter_collection_.leader.open(scope.as_cpu());
+            }
+            else
+            {
+                counter_leader_ = counter_collection_.leader.open(scope.as_thread());
+            }
```
bmario, line 68: overload

include/lo2s/perf/syscall/reader.hpp
```diff
@@ -68,110 +69,50 @@ class Reader : public EventReader<T>
...
+
+        exit_ev_.set_output(enter_ev_);
+
+        if (!config().syscall_filter.empty())
         {
-            close(other_fd_);
-            close(fd_);
-            throw;
+            enter_ev_.set_syscall_filter();
+            exit_ev_.set_syscall_filter();
```
bmario, line 97: Add an early exit for an empty input in set_syscall_filter() and make this call here just unconditional.

include/lo2s/perf/reader.hpp
```diff
@@ -0,0 +1,368 @@
...
+{
+public:
+    RawEvent(bool enable_on_exec);
+};
+
+/**
+ * Contains an opened instance of PerfEvent.
+ * Use PerfEvent.open() to construct an object
+ */
+class PerfEventInstance
```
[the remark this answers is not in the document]
bmario: Also, I have the feeling, that it might be good to have a template argument to this class, which is the actual event used, e.g., `EventGuard<TracepointEvent> event_;`.

bmario (changes requested):

src/monitor/tracepoint_monitor.cpp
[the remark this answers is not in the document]
teto519f: I'm tempted to either add a init_attrs_with_config() function, which sets all of them at once, or a init_with_config_or_else() type of function, where it tries to use config() but if it fails just sets default values. Is one of them alright?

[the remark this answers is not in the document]
cvonelm: I personally dont like pulling the config() into the *Event in any way.

include/lo2s/perf/reader.hpp
```diff
@@ -0,0 +1,368 @@
...
+
+    bool is_valid() const
+    {
+        return fd_ >= 0;
+    };
+
+    template <class T>
+    T read()
+    {
+        T val;
```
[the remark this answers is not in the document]
teto519f: `static_assert(std::is_pod_v<T> == true);` will fail, as UserspaceReadFormat has a user-defined constructor.

src/perf/reader.cpp
```diff
@@ -0,0 +1,463 @@
...
+    }
+    return cpus_;
+}
+
+bool PredefEvent::is_valid() const
+{
+    return (availability_ != Availability::UNAVAILABLE);
+}
+
+void PredefEvent::set_availability()
```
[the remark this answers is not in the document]
teto519f: The function can't be private, since it is needed in perf/event_provider.cpp: event_is_openable()

include/lo2s/perf/reader.hpp
teto519f: There are multiple ways to get cpus. Once through `Topology::instance().cpus()`, which takes cpus from `/sys/devices/system/cpu/online` and one which takes them from `/sys/bus/event_source/devices/cpus` or `-/cpumask`. 
It doesn't look like they can be combined.

cvonelm: Use Topology::instance().cpus() when cpus or cpumask do not exist.

Use either cpus or cpumask when they do exist. An event cant have both a cpus and cpumask file.


Thie logic here is that cpus and cpumask specify that some event is not available on all CPUs and instead is only available on the cpus listed in that file (why there needs to be two different files, nobody knows). 

If neither file exist we expect the event to be available on every cpu, so use the list of all cpus from Topology.

include/lo2s/perf/reader.hpp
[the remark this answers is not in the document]
cvonelm: Remove the constructor. If I see it correctly, the only thing it does is to set the members to 0. This can be ensured by different ways (memset) if there are places where this might cause issues

include/lo2s/perf/reader.hpp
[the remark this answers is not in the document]
teto519f: `PredefEvent` is not needed anymore. Should `PerfEvent` still be renamed to `Event`?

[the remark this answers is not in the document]
cvonelm: I think so. This is already in the perf namespace so it being a perf::PerfEvent instead of a perf::Event seems a bit redundant.

include/lo2s/perf/reader.hpp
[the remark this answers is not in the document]
teto519f: But the type of `Event` in `EventGuard` isn't being checked anywhere. Is there even a use for this information in `EventGuard`?

include/lo2s/measurement_scope.hpp
```diff
@@ -111,6 +117,8 @@ struct MeasurementScope
             return fmt::format("block layer I/O events for {}", scope.name());
         case MeasurementScopeType::SYSCALL:
             return fmt::format("syscall events for {}", scope.name());
+        case MeasurementScopeType::TRACEPOINT:
+            return fmt::format("tracepoints for {}", scope.name());
```
cvonelm, line 121: tracepoints -> tracepoint events

include/lo2s/perf/bio/writer.hpp
```diff
@@ -182,9 +184,11 @@ class Writer
     trace::Trace& trace_;
     time::Converter& time_converter_;
 
-    tracepoint::EventFormat bio_queue_;
-    tracepoint::EventFormat bio_issue_;
-    tracepoint::EventFormat bio_complete_;
+    // Unabailable until get_tracepoints() is called
```
cvonelm, line 187: Unabailable -> Unavailable

include/lo2s/perf/event.hpp
```diff
@@ -0,0 +1,288 @@
...
+
+    /**
+     * opens child as a counter of the calling (leader) event
+     */
+    EventGuard open_child(Event child, ExecutionScope location, int cgroup_fd = -1);
+
+    void enable();
+    void disable();
+
+    void set_id(const uint64_t& id)
```
cvonelm, line 245: This ioctl does not set the id, it  gets the id. From the `perf_event_open` manpage:

> This returns the event ID value for the given event file descriptor.  

please rewrite as uint64_t get_id() function

src/perf/event.cpp
```diff
@@ -0,0 +1,658 @@
...
+
+constexpr std::uint64_t operator"" _u64(unsigned long long int lit)
+{
+    return static_cast<std::uint64_t>(lit);
+}
+
+constexpr std::uint64_t bit(int bitnumber)
+{
+    return static_cast<std::uint64_t>(1_u64 << bitnumber);
+}
```
bmario, line 49: This is probably the most convoluted code to obfuscate `(1ull << bitnumber)`

src/perf/event.cpp
```diff
@@ -0,0 +1,658 @@
...
+        BM_WHOLE_MATCH,
+        BM_BEGIN,
+        BM_END,
+    };
+
+    std::uint64_t mask = 0x0;
+
+    static const std::regex bit_mask_regex(R"((\d+)?(?:-(\d+)))");
+    const std::sregex_iterator end;
+    std::smatch bit_mask_match;
```
bmario, line 86:
```suggestion
```

src/perf/event.cpp
```diff
@@ -0,0 +1,658 @@
...
+}
+
+EventGuard::EventGuard(Event& ev, std::variant<Cpu, Thread> location, int group_fd, int cgroup_fd)
+: ev_(ev)
+{
+    // can be deleted when scope gets replaced
+    ExecutionScope scope;
+    std::visit(overloaded{ [&](Cpu cpu) { scope = cpu.as_scope(); },
+                           [&](Thread thread) { scope = thread.as_scope(); } },
+               location);
```
bmario, line 597:
```suggestion
    std::visit([&scope](auto loc) { scope = loc.as_scope(); },
               location);
```

`overloaded` is not needed in this case. Just use a simply lambda instead.

src/perf/event.cpp
```diff
@@ -0,0 +1,658 @@
...
+    }
+}
+
+void EventGuard::enable()
+{
+    auto ret = ioctl(fd_, PERF_EVENT_IOC_ENABLE);
+    if (ret == -1)
+    {
+        throw_errno();
+    }
```
bmario, line 619: I think `ret` is is not necessary here.
```suggestion
    if (ioctl(fd_, PERF_EVENT_IOC_ENABLE) == -1)
    {
        throw_errno();
    }
```

src/perf/event.cpp
```diff
@@ -0,0 +1,658 @@
...
+}
+
+void EventGuard::disable()
+{
+    auto ret = ioctl(fd_, PERF_EVENT_IOC_DISABLE);
```
bmario, line 624: again, `ret` not required.

src/perf/event_provider.cpp
```diff
@@ -687,23 +350,109 @@ bool EventProvider::has_event(const std::string& name)
...
+
+tracepoint::TracepointEvent EventProvider::create_raw_tracepoint_event(const std::string& name,
+                                                                       const bool& enable_on_exec)
+{
+    tracepoint::TracepointEvent event(name, enable_on_exec);
+    event.set_sample_period(0);
+    apply_default_attrs(event);
+
+    return event;
+}
```
bmario, line 389: these two functions are identical, but the name.

src/perf/event_provider.cpp
```diff
@@ -687,23 +350,109 @@ bool EventProvider::has_event(const std::string& name)
...
-            events.push_back(event.second.description);
+            events.push_back(std::move(event.second));
         }
     }
 
     return events;
 }
+
+tracepoint::TracepointEvent EventProvider::create_tracepoint_event(const std::string& name,
+                                                                   const bool& enable_on_exec)
```
bmario, line 372:
```suggestion
                                                                   bool enable_on_exec)
```

src/perf/event_provider.cpp
```diff
@@ -687,23 +350,109 @@ bool EventProvider::has_event(const std::string& name)
...
+    Event event(name, type, config, config1);
+    event.set_sample_period(0);
+
+    apply_config_attrs(event);
+    event.set_exclude_kernel(true); // overwrite config value
+
+    return event;
+}
+
+SysfsEvent EventProvider::create_sampling_event(const bool& enable_on_exec)
```
bmario, line 414:
```suggestion
SysfsEvent EventProvider::create_sampling_event(bool enable_on_exec)
```

src/perf/event_provider.cpp
```diff
@@ -687,23 +350,109 @@ bool EventProvider::has_event(const std::string& name)
...
+}
+
+SysfsEvent EventProvider::create_sysfs_event(const std::string& name)
+{
+    SysfsEvent event(name);
+    event.set_sample_period(0);
+    apply_config_attrs(event);
+
+    return event;
+}
```
bmario, line 441: These two functions are identical

src/perf/event_provider.cpp
```diff
@@ -235,20 +171,23 @@ static void populate_event_map(EventProvider::EventMap& map)
...
-                name_fmt.str(), PERF_TYPE_HW_CACHE,
-                make_cache_config(cache.id, operation.id.op_id, operation.id.result_id));
+            // can't use create_raw_event here, will create deadlock
+            Event event(name_fmt.str(), PERF_TYPE_HW_CACHE,
+                        make_cache_config(cache.id, operation.id.op_id, operation.id.result_id));
+
+            event.set_sample_period(0);
+            event.set_exclude_kernel(true);
+            event.set_watermark(16);
+            event.set_clock_attrs(true, CLOCK_MONOTONIC_RAW);
```
bmario, line 181: use  apply_default_attrs` ?

src/perf/time/reader.cpp
```diff
@@ -55,39 +55,15 @@ Reader::Reader()
...
-    attr.bp_len = HW_BREAKPOINT_LEN_8;
-    attr.wakeup_events = 1;
-    attr.sample_period = 1;
-#else
-    attr.type = PERF_TYPE_HARDWARE;
-    attr.config = PERF_COUNT_HW_INSTRUCTIONS;
-    attr.sample_period = 100000000;
-    attr.task = 1;
-#endif
+    Event event = EventProvider::instance().create_time_event((uint64_t)&local_time);
```
bmario, line 59: Don't use C-Style casts.

include/lo2s/perf/event.hpp
```diff
@@ -0,0 +1,291 @@
...
+
+    friend bool operator<(const Event& lhs, const Event& rhs)
+    {
+        return memcmp(&lhs.attr_, &rhs.attr_, sizeof(struct perf_event_attr));
+    }
+
+    friend bool operator>(const Event& lhs, const Event& rhs)
+    {
+        return memcmp(&lhs.attr_, &rhs.attr_, sizeof(struct perf_event_attr));
+    }
```
bmario, line 179: these are identical, so two Event instances a and b are equal: a == b, or both
a < b and a > b.

include/lo2s/perf/event.hpp
```diff
@@ -0,0 +1,291 @@
...
+    EventGuard();
+    EventGuard(Event& ev, std::variant<Cpu, Thread> location, int group_fd, int cgroup_fd);
+
+    EventGuard(EventGuard&) = delete;
+    EventGuard& operator=(const EventGuard&) = delete;
+
+    EventGuard(EventGuard&& other)
+    {
+        std::swap(fd_, other.fd_);
+        std::swap(ev_, other.ev_);
```
bmario, line 227: This doesn't work. fd_ and ev_ are not initialized.

include/lo2s/perf/event.hpp
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    ~EventGuard()
+    {
+        close(fd_);
+    }
+
+protected:
+    int fd_;
+    Event ev_;
```
bmario, line 287: You can't make a copy of Event. In fact, I'm not sure, if you even need a reference or pointer to event at all.

include/lo2s/perf/event.hpp
```diff
@@ -0,0 +1,291 @@
...
+     */
+    EventGuard open_child(Event child, ExecutionScope location, int cgroup_fd = -1);
+
+    void enable();
+    void disable();
+
+    uint64_t get_id()
+    {
+        uint64_t id;
+        ioctl(fd_, PERF_EVENT_IOC_ID, &id);
```
bmario, line 248: check return value for errors

include/lo2s/perf/event.hpp
```diff
@@ -0,0 +1,291 @@
...
+};
+
+/**
+ * Contains an opened instance of Event.
+ * Use any Event.open() method to construct an object
+ */
+class EventGuard
+{
+public:
+    EventGuard();
```
bmario, line 218: There should not be default constructed EventGuard instances. the default constructor should be ` = delete`.

src/perf/event.cpp
```diff
@@ -0,0 +1,658 @@
...
+    memset(&attr_, 0, sizeof(attr_));
+    attr_.size = sizeof(attr_);
+    attr_.type = PERF_TYPE_RAW;
+
+    attr_.config = 0;
+    attr_.config1 = 0;
+
+    attr_.sample_period = 0;
+    attr_.exclude_kernel = 1;
+}
```
bmario, line 144: I don't think a default constructed event should be possible.

src/perf/event.cpp
```diff
@@ -0,0 +1,658 @@
...
+
+Event::Event([[maybe_unused]] uint64_t addr, bool enable_on_exec)
```
bmario, line 146: this does not set the `name_` I don't want Event instances without a name.
The real issue is that this constructor is used for some special task and should better be a factory function

src/perf/event.cpp
```diff
@@ -0,0 +1,658 @@
...
+    attr_.wakeup_events = 1;
+#else
+    attr_.type = PERF_TYPE_HARDWARE;
+    attr_.config = PERF_COUNT_HW_INSTRUCTIONS;
+    attr_.sample_period = 100000000;
+    attr_.task = 1;
+#endif
+}
+
+Event::Event(std::string name, perf_type_id type, std::uint64_t config, std::uint64_t config1)
```
bmario, line 164:
```suggestion
Event::Event(const std::string& name, perf_type_id type, std::uint64_t config, std::uint64_t config1)
```

src/perf/event.cpp
```diff
@@ -0,0 +1,658 @@
...
+        else
+            availability_ = Availability::SYSTEM_MODE;
+    }
+    else
+    {
+        if (sys_ev.get_fd() == -1)
+            availability_ = Availability::PROCESS_MODE;
+        else
+            availability_ = Availability::UNIVERSAL;
+    }
```
bmario, line 370: Make use of the fact, that `Availability` can be used like a bitfield.

```
availability_ = Availability::UNAVAILABLE;
if (proc_ev.get_fd() != -1) availability_ |= Availability::PROCESS_MODE;
if (sys_ev.get_fd() != -1) availability_ |= Availability::SYSTEM_MODE;

File: src/perf/event.cpp [hunk 48]
```diff
@@ -0,0 +1,658 @@
...
+        ED_PMU,
+        ED_NAME,
+    };
+
+    parse_pmu_path(ev_name);
+
+    Log::debug() << "parsing event description: pmu='" << pmu_name_ << "', event='" << name_ << "'";
+
+    // read PMU type id
+    std::underlying_type<perf_type_id>::type type = read_file_or_else(pmu_path_ / "type", 0);
```
bmario, line 421: This `read_file_or_else` business is very strange. You're making a function, that reads a file, but ignores errors and returns a sentinel value instead. Then you check for the same sentinel value and throw an error.

src/perf/event_provider.cpp
teto519f: They are not, as one uses config() and the other does not. 
But I agree, they are pretty similar. Will merge them into one.

src/perf/event_provider.cpp
teto519f: Same situation as above.

src/perf/event.cpp
teto519f: Which alternative would you suggest? Copying the ifstream code from read_file_or_else every time or making some try_read_file function, which would also copy the code? 
I agree that this kinda alienates the function, but at the same time it avoids code duplication.

src/perf/event.cpp
teto519f: Which name should a time event get? Or in other words: From where should the event take its name?

src/perf/event_provider.cpp
bmario: Then add another EventProvider "factory" for the whole thing?

src/perf/event.cpp
bmario: The issue I see here is that it feels like you're trying to import error handling from a different language into C++. `X_or_else` feels very rusty. A more idiomatic way to implement `read_file` would be to either return a `std::optional` or throw an exception.

src/perf/event.cpp
teto519f: Why not? This way you can say that a reader holds an event without needing to know the specifics.

bmario: Because of the unknown specifics, you end up with an object that cannot be used. It's not initialized, a constructor should never give an ususable object.

src/perf/event.cpp
bmario: "Time" sounds as good to me as anything else.

bmario (approved):

PR #341, opened 2024-08-02
feat: Add C++ Mmap Wrapper

cvonelm:
In preperation for CUPTI support, this commit adds a SharedMemory class that wraps Mmap and converts the EventReader to use it.

### DIFF
--- a/include/lo2s/shared_memory.hpp
+++ b/include/lo2s/shared_memory.hpp
@@ -1,0 +1,119 @@
+/*
+ * This file is part of the lo2s software.
+ * Linux OTF2 sampling
+ *
+ * Copyright (c) 2016,
+ *    Technische Universitaet Dresden, Germany
+ *
+ * lo2s is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * lo2s is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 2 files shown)


bmario (changes requested):
I added comments in #294

include/lo2s/shared_memory.hpp
```diff
@@ -0,0 +1,109 @@
...
+        addr_ = other.addr_;
+        size_ = other.size_;
+        other.addr_ = nullptr;
+    }
+
+    SharedMemory& operator=(SharedMemory&& other)
+    {
+        addr_ = other.addr_;
+        size_ = other.size_;
+        other.addr_ = nullptr;
```
bmario, line 57: This is not correct behavior. Move-Assignment has to work for two "valid" objects. You can't just simply discard one mapping.

Using std::swap was one correct implementation. Another correct solution would be to `munmap(addr_)`  if it is not the `nullptr` before overwriting it with `other.addr_`.

include/lo2s/shared_memory.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+    template <typename T>
+    const T* as() const
+    {
+        return reinterpret_cast<const T*>(addr_);
+    }
+
+    ~SharedMemory()
+    {
+        if (addr_ != nullptr)
```
bmario, line 102: extract into function

include/lo2s/shared_memory.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+        other.addr_ = nullptr;
+    }
+
+    SharedMemory& operator=(SharedMemory&& other)
+    {
+        if (addr_ != nullptr)
+        {
+            munmap(addr_, size_);
+        }
```
bmario, line 59: extract into function and use in desctructor

bmario (changes requested):

bmario (approved):

PR #348, opened 2024-10-29
fix(cmake): Fix static build

cvonelm:
Some libraries were not correctly (or at all) checking for static variants if lo2s_USE_STATIC_LIBS was requested, so I fixed that.

### DIFF
--- a/cmake/FindAudit.cmake
+++ b/cmake/FindAudit.cmake
@@ -1,0 +1,54 @@
+# Copyright (c) 2017, Technische Universität Dresden, Germany
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are permitted
+# provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list of conditions
+#    and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions
+#    and the following disclaimer in the documentation and/or other materials provided with the
+#    distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse
+#    or promote products derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
+# IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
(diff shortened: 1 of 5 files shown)


cmake/FindAudit.cmake
```diff
@@ -0,0 +1,51 @@
...
+        if(Audit_USE_STATIC_LIBS)
+            try_compile(BUILD_AUDIT
+                SOURCES "${CMAKE_CURRENT_LIST_DIR}/test_static_audit.c"
+                LINK_LIBRARIES ${Audit_STATIC_LIBRARIES}
+                COMPILE_DEFINITIONS "-static")
+            if(NOT BUILD_AUDIT)
+                set(Audit_FOUND OFF CACHE INTERNAL "")
+            endif()
+        endif()
+        add_library(Audit::Audit ALIAS PkgConfig::Audit)
```
bmario, line 49: I think you have to define your own target if static should be used. I'd assume PkgConfig::Audit always links the shared lib?

cmake/FindRadare.cmake
```diff
@@ -0,0 +1,52 @@
...
+        if(Radare_USE_STATIC_LIBS)
+            try_compile(BUILD_RADARE
+                SOURCES "${CMAKE_CURRENT_LIST_DIR}/test_static_radare.c"
+                LINK_LIBRARIES ${Radare_STATIC_LIBRARIES}
+                COMPILE_DEFINITIONS "-static")
+            if(NOT BUILD_RADARE)
+                set(Radare_FOUND OFF CACHE INTERNAL "")
+            endif()
+        endif()
+        add_library(Radare::Radare ALIAS PkgConfig::Radare)
```
bmario, line 50: Same as with the audit target. you likely need your own target to link against the static libs.

bmario (changes requested):

cmake/FindRadare.cmake
```diff
@@ -0,0 +1,59 @@
...
+
+include(${CMAKE_CURRENT_LIST_DIR}/UnsetIfUpdated.cmake)
+
+option(Radare_USE_STATIC_LIBS "Link Radare statically" OFF)
+
+UnsetIfUpdated(Radare_LIBRARIES Radare_USE_STATIC_LIBS)
+
+if(PkgConfig_FOUND)
+    # Breaking API changes in 5.8.0
+    pkg_check_modules(Radare IMPORTED_TARGET r_main>=5.8.0)
```
bmario, line 36: This thing is now needed
https://github.com/tud-zih-energy/lo2s/pull/355/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR118

bmario (changes requested):

bmario (approved):

PR #358, opened 2025-01-22
Fix various remaining issues with EventDescription Rework, streamline interfaces, move configuration of Events in own EventConfig class.

cvonelm:
#Move Event Configuration to EventConfig class
    
    Designwise, the EventConfig class should handle all the ugly details of
    configuring the Events from the given Config(), such as setting specific
    bits.
    
    The sample::Reader for example asks the EventConfig for a sample event.
    The EventConfig then selects the correct sampling event using the
    EventProvider and the Config, configures all the options, such as
    sample_period and then gives it back to the sample::Reader.
    The sample::Reader does not need to change anything about the event.
    
    The EventProvider should only provide events, reading them from
    different sources, such as pre-defined, PFM, PMU events.
    
    The CounterProvider did not do anything special enough to warrant its
    continuing existence, merge it with EventConfig
    
    Smaller fixes
    - delete needless getter in CounterCollection
    - provide default values for bools in Config()
    - check exclude_kernel once in EventConfig() instead of for every event

 # Test for debugfs when specifying tracepoint dependent options,
    print [CPUS 0-0] as [CPU 0]

#   Refactor Event class
    
    - Make Event clean base class. Use SimpleEvent when the  event configuration needs to
      be specified manually.
    - Move almost all logic outside of the Event class. Event Configuration
      should happen fully from the outside
    - fix boolean operators
    - move sysfs specific code from Event to SysfsEvent

# Pass TracepointEvent directly instead of name of event
    
    Also move use-case specific tracepoint code out of TracepointEvent

# Add definition of base_path into read_proc()
    
    Somehow, base_path does not get initialized correctly before read_proc
    is called, so move it all into read_proc


include/lo2s/perf/event.hpp
```diff
@@ -59,21 +60,29 @@ enum class Availability
...
 /**
  * Base class for all Event types
  * contains common attributes
  */
+
 class Event
 {
 public:
-    Event(const std::string& name, perf_type_id type, std::uint64_t config,
+    Event(const std::string name, perf_type_id type, std::uint64_t config,
```
bmario, line 85:
```suggestion
    Event(const std::string& name, perf_type_id type, std::uint64_t config,
```

include/lo2s/perf/event.hpp
```diff
@@ -132,82 +136,187 @@ class Event
...
     std::set<Cpu> cpus_;
     Availability availability_ = Availability::UNAVAILABLE;
+};
 
-    std::filesystem::path pmu_path_;
-    std::string pmu_name_;
+class SimpleEvent : public Event
+{
+public:
+    SimpleEvent(const std::string name, perf_type_id type, std::uint64_t config,
```
bmario, line 300:
```suggestion
    SimpleEvent(const std::string& name, perf_type_id type, std::uint64_t config,
```

include/lo2s/perf/event_config.hpp
```diff
@@ -0,0 +1,70 @@
...
+    {
+        static EventConfig e;
+        return e;
+    }
+
+    counter::CounterCollection counters_for(MeasurementScope scope);
+
+    Event create_time_event(uint64_t local_time);
+    Event create_sampling_event();
+    perf::tracepoint::TracepointEvent create_tracepoint_event(std::string name);
```
bmario, line 49:
```suggestion
    perf::tracepoint::TracepointEvent create_tracepoint_event(const std::string& name);
```

include/lo2s/perf/event_config.hpp
```diff
@@ -0,0 +1,70 @@
...
+    std::vector<perf::tracepoint::TracepointEvent> get_tracepoints();
+
+private:
+    // When we poll on the fd given by perf_event_open, wakeup, when our buffer is 80% full
+    // Default behaviour is to wakeup on every event, which is horrible performance wise
+    void watermark(Event& ev)
+    {
+        ev.set_watermark(0.8 * config().mmap_pages * sysconf(_SC_PAGESIZE));
+    }
```
bmario, line 58: Why is this here? It seems to be not worth a function on its own. Especially because the name is not really telling.

include/lo2s/perf/event.hpp
```diff
@@ -132,82 +136,187 @@ class Event
         unit_ = unit;
     }
 
-    void clock_attrs([[maybe_unused]] bool use_clockid, [[maybe_unused]] clockid_t clockid)
+    void set_clockid(int64_t clockid)
```
bmario, line 139: Use `optional<clockid_t>`

include/lo2s/perf/event_config.hpp
```diff
@@ -0,0 +1,70 @@
...
+        static EventConfig e;
+        return e;
+    }
+
+    counter::CounterCollection counters_for(MeasurementScope scope);
+
+    Event create_time_event(uint64_t local_time);
+    Event create_sampling_event();
+    perf::tracepoint::TracepointEvent create_tracepoint_event(std::string name);
+    std::vector<perf::tracepoint::TracepointEvent> get_tracepoints();
```
bmario, line 50: This feels super dirty. Why copy?

bmario (commented):
I think you finally lost me with the naming.

What do the Event instances represent? (And their subclass instances)
What is the difference between EventProvider and EventConfig?

include/lo2s/config.hpp
```diff
@@ -83,23 +83,26 @@ struct Config
...
     bool use_syscalls = false;
     std::vector<int64_t> syscall_filter;
     // NEC SX-Aurora Tsubasa
-    bool use_nec;
+    bool use_nec = false;
     std::chrono::microseconds nec_read_interval;
     std::chrono::milliseconds nec_check_interval;
     // Nvidia CUPTI
-    bool use_nvidia;
+    bool use_nvidia = false;
```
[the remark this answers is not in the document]
cvonelm: Atleast with use_nvidia, I think, I had a issue that it came out as true, because it was set nowhere

include/lo2s/perf/pfm.hpp
```diff
@@ -72,12 +72,7 @@ class PFM4
             return std::nullopt;
         }
 
-        Event ev = Event(ev_desc, (perf_type_id)attr.type, attr.config, attr.config1);
-
-        if (!ev.event_is_openable())
-        {
-            return std::nullopt;
-        }
+        Event ev = SimpleEvent(ev_desc, (perf_type_id)attr.type, attr.config, attr.config1);
```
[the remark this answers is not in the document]
cvonelm: None of the EventAttr derivations have their own fields, so this is not an issue.

TracepointEventAttr has its own fields but it has a complete separate code path for event creation that return TracepointEventAttrs

include/lo2s/perf/event_config.hpp
cvonelm: We use this calculation for the watermark at three different spots in the code and I dont want to spread the arcane knowledge that watermark is 80% of config().mmap_pages * page_size all through the code.

src/perf/event_attr.cpp
```diff
@@ -42,6 +41,23 @@ namespace lo2s
 namespace perf
 {
 
+std::set<Cpu> test_cpus(EventAttr ev)
```
bmario, line 44: Please give it a better name. What does it test for?

include/lo2s/perf/event_attr.hpp
```diff
@@ -59,22 +60,39 @@ enum class Availability
     UNIVERSAL
 };
 
-inline Availability& operator|=(Availability& a, Availability b) noexcept
+enum class EventFlag
```
bmario, line 63: While I liked this at first, In practice it just feels bad. There is this setFlags method taking a whole bunch of flags as parameters, but is always used only with one flag. Then again, all these flags aren't flags, but decode some manipulation to internal attributes. All flags are bunched up in one enum, even though they have nothing to do with each other. Please use member functions instead. And add documentation to the methods, how each of the methods manipulates the event to a certain behavior.

src/monitor/process_monitor.cpp
```diff
@@ -51,15 +50,20 @@ void ProcessMonitor::insert_thread(Process process, Thread thread, std::string n
...
+    ExecutionScope scope = ExecutionScope(thread);
     if (config().sampling ||
-        perf::counter::CounterProvider::instance().has_group_counters(ExecutionScope(thread)) ||
-        perf::counter::CounterProvider::instance().has_userspace_counters(ExecutionScope(thread)))
+        !perf::EventComposer::instance()
+             .counters_for(MeasurementScope::group_metric(scope))
+             .counters.empty() ||
+        !perf::EventComposer::instance()
+             .counters_for(MeasurementScope::userspace_metric(scope))
+             .counters.empty())
```
bmario, line 60: please add a "has_counters_for" method to the EventComposer class.

src/monitor/scope_monitor.cpp
```diff
@@ -56,14 +56,18 @@ ScopeMonitor::ScopeMonitor(ExecutionScope scope, MainMonitor& parent, bool enabl
         add_fd(syscall_writer_->fd());
     }
 
-    if (perf::counter::CounterProvider::instance().has_group_counters(scope))
+    if (!perf::EventComposer::instance()
+             .counters_for(MeasurementScope::group_metric(scope))
+             .counters.empty())
```
bmario, line 61: has_counters_for

src/monitor/scope_monitor.cpp
```diff
@@ -56,14 +56,18 @@ ScopeMonitor::ScopeMonitor(ExecutionScope scope, MainMonitor& parent, bool enabl
...
     {
         group_counter_writer_ =
             std::make_unique<perf::counter::group::Writer>(scope, parent.trace(), enable_on_exec);
         add_fd(group_counter_writer_->fd());
     }
 
-    if (perf::counter::CounterProvider::instance().has_userspace_counters(scope))
+    if (!perf::EventComposer::instance()
+             .counters_for(MeasurementScope::userspace_metric(scope))
+             .counters.empty())
```
bmario, line 70: has_counters_for

src/perf/counter/userspace/reader.cpp
```diff
@@ -46,42 +46,15 @@ namespace userspace
...
-                Log::error() << "perf_event_open for counter failed";
-                throw_errno();
-            }
-            else
-            {
-                counters_.emplace_back(std::move(counter.value()));
-            }
-        }
+        std::optional<EventGuard> counter = event.open(scope);
+        counters_.emplace_back(std::move(counter.value()));
```
bmario, line 57:
```suggestion
        counters_.emplace_back(event.open(scope));
```

bmario (changes requested):

PR #423, opened 2026-09-21
fix(dwarf/symbols): Fix Symbol Resolution for static binaries.

### LINKED ISSUE #421: Debug symbols not read / no functions
When using lo2s with NAS Parallel Benchmarks, only the binary information is stored in the trace, no function that is entered.
Tested systems:
- slurm-docker-cluster 25.10 image (gcc)
- Ubuntu 25.10
- TU Dresden HPC system Barnard (Red Hat Enterprise Linux 9.6)

This occurs also with lo2s 1.8.0 while lo2s 1.7.0 was tested on barnard and works. Maybe it is related to "binutils has been replaced with elfutils, hopefully providing better symbol resolution" (quote from the 1.8.0 release)

cvonelm:
The symbol table of the application contains a list of (among other things) all functions in the program.

All entries in the symbol table have a value. This value specifies the address of the symbol.

For relocatable objects, this address is relative to the offset at which the application is mapped:

program_counter - mapping offset -> address in symbol table

For static objects, the address is absolute:

program_counter -> address in symbol table.

Which is all fine and dandy, HOWEVER:

WHY COULDNT THEY JUST DEFINE the mapping offset to be ZERO in the case of static objects? Then the same code would work for BOTH relocatable and static objects.

Instead, the mapping offset at which the binary is mapped is NOT ZERO in the static case and you are just meant to remember that if the thing is not relocatable, then you don't need to subtract the offset.

dammit.

Anyways, this commit adds the check for relocatibility ranted about above and provides better debug output to trace how symbols are resolved.

## Checklist

- [x] I have checked how my changes affect lo2s CLI options and changed those accordingly

This fixes #421


cvonelm:
@rschoene On romeo with ifx/icx, this atleast fixed the issue with no symbols being resolved, now it looks good:

```
CALLING_CONTEXT_ENTER                          0      544639189348214  Calling Context: "bt.B.x (2794847)" <1>, Unwind Distance: 2
METRIC                                         0      544639189348214  Metric: 1, 1 Value: ("CPU" <0>; INT64; 105)
CALLING_CONTEXT_SAMPLE                         0      544639189523174  Calling Context: "compute_rhs_.DIR.OMP.PARALLEL.25@bt.B.x:1" <546>, Unwind Distance: 2, Interrupt Generator: "perf HW_INSTRUCTIONS" <0>
CALLING_CONTEXT_SAMPLE                         0      544639190575684  Calling Context: "compute_rhs_.DIR.OMP.PARALLEL.25@bt.B.x:1" <636>, Unwind Distance: 2, Interrupt Generator: "perf HW_INSTRUCTIONS" <0>
CALLING_CONTEXT_SAMPLE                         0      544639191621954  Calling Context: "compute_rhs_.DIR.OMP.PARALLEL.25@bt.B.x:1" <535>, Unwind Distance: 2, Interrupt Generator: "perf HW_INSTRUCTIONS" <0>
CALLING_CONTEXT_LEAVE                          0      544639192325374  Calling Context: "bt.B.x (2794847)" <1>
```

